PR #40, opened 2015-02-08
Fix small typo in README

mize:
Looks like ref to `local` accidentally made it's way through during latest README edits.


nolanlawson:
cool, thanks!

PR #43, opened 2015-03-05
Changes watcher improvements

rsutphin:
Style, error-avoidance, and performance fixes for the automatic changes watcher. There are details in the commits; at a high-level:

* Don't throw exceptions when there's a change to a design doc or other non-relational_pouch doc.
* Don't keep changes around in memory since the watcher doesn't need them.
* Don't load records into the store that aren't already there. This makes a huge speed and memory difference on large databases.

All of these changes were initially suggested & implemented by my colleague @moses.

### DIFF
--- a/lib/pouchdb-adapter.js
+++ b/lib/pouchdb-adapter.js
@@ -15,29 +15,31 @@ export default DS.RESTAdapter.extend({
       live: true,
       returnDocs: false
     }).on('change', function (change) {
-      var obj = this.db.rel.parseDocID(change.id);
-      // skip changes for non-relational_pouch docs. E.g., design docs.
-      if (!obj.type || !obj.id || obj.type === '') { return; }
+      Ember.run(function () {
+        var obj = this.db.rel.parseDocID(change.id);
+        // skip changes for non-relational_pouch docs. E.g., design docs.
+        if (!obj.type || !obj.id || obj.type === '') { return; }
 
-      var store = this.container.lookup('store:main');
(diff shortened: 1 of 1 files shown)


rsutphin:
I left the PR in separate commits so the changes are easier to review. I'm happy to squash them before merging if you'd prefer.

nolanlawson:
Loving all these changes, and the separate commits makes it a lot easier to understand what's going on. :+1: 

The only concern I have is that I see the complexity of this project growing, while we still aren't running any tests. :( All the tests are isolated to relational-pouch, because I wasn't familiar enough with Ember to set up a proper test suite.

That can wait for a future commit, though. For now, this looks great, thanks!

lokeshj:
What if I do want a model to be loaded into the store if it weren't already? e.g. in-app notifications.

nolanlawson:
@lokeshj Sounds like an interesting solution. Could you document in in the README? :D

lokeshj:
sure, will submit my solution

rsutphin:
Overriding will sort of work, though wouldn't you'd have to copy the entire existing method to keep the existing behavior? Sounds hard to maintain. 

If we want to make it documented that you can override it, we should make it not private (the `_` prefix convention suggests it's private). 

Alternative: how about a second change listener (defined in your application) that just reacts to new records?

lokeshj:
@rsutphin yeah that sounds much better. I did have to copy the entire method and it certainly is tough to maintain.

PR #63, opened 2015-06-03
Replace typeKey by modelName

OleRoel:
This is a PR for issue https://github.com/nolanlawson/ember-pouch/issues/62


nolanlawson:
Hm, this looks fine to me, but I'm worried that if I merge this, then ember-pouch will break for people on older versions of Ember Data. (And by the way, I have repeated that sentence about 5 times for various PRs in this project; Ember Data changes fast... :stuck_out_tongue:)

Can someone else please take a look at this PR and confirm that it'll be okay? Do we need to bump the major version to indicate a breaking change?

OleRoel:
What's about a new branch which will keep pace with Ember Data and fixes upcomming deprecation warnings and code breaks? This can then be merged back to master when Ember finaly (in the near future?) becomes 1.0. The branch then doesn't need to take care of backward compatibility. A drawback is the extra burden to integrate changes from master to the 'shadow' branch whenever they occur.

nolanlawson:
Hm, that sounds like a handful to me. I would much prefer to just integrate your PR and then call it a breaking change if it is truly a breaking change (i.e. release 2.0).

OleRoel:
@nolanlawson We should then also remove the backwards compatibility here when the PR is integrated into the master branch:

https://github.com/nolanlawson/ember-pouch/blob/master/addon/adapters/pouch.js#L136

(Btw: Ember Data 1.0 is coming soon since Mar 18, 2014: http://emberjs.com/blog/2014/03/18/the-road-to-ember-data-1-0.html)

rsutphin:
To be clear, this it not just a backwards incompatible change in the sense that it won't work with older versions of ember-data. `modelName` is not just another name for `typeKey` — it will have different values for a model whose name has more than one word in it:

> the `modelName` is a dasherized string. For example, if you had a model called `TacoShop`, it would be stored on the model's constructor's `modelName` property as `taco-shop`, whereas previously it would be stored as `tacoShop`.

([source](http://emberjs.com/blog/2015/05/21/ember-data-1-0-beta-18-released.html))

This will require a substantial data migration for any existing application using ember-pouch. I would be strongly in favor of _not_ making this change for that reason, and instead replacing direct uses of `typeKey` with a function which returns a camelized version of `modelName`.

(If you're doing that already, it would also be pretty straightforward to use `typeKey` if `modelName` isn't available and not lose backwards compatibility immediately. In the future when we want to drop support for the ember-data betas, this would make it easier to do all at once, instead of with a major version bump every time ember-data does something backwards-incompatible.)

broerse:
The new Ember CLI 0.2.7 starts warning: `DEPRECATION: Usage of `typeKey` has been deprecated and will be removed in Ember Data 1.0. It has been replaced by `modelName` on the model class.` I vote for integrating this PR and call it a breaking change.

nolanlawson:
As @rsutphin points out, though, won't this break for people who named their models `CamelCaseStyle`?

OleRoel:
@rsutphin Thank you for raising an important concern here. Are you suggesting a getter function like

```javascript
getModelName(type) {
  if (typeof type.modelName !== 'undefined') {
    return Em.String.camelize(type.modelName)
  } else {
    return type.typeKey
  }
}
```
which will replace all occurances of type.typeKey? A `TestModel` will be transformed to `test-model` in Ember-CLI with help of `ember generate model TestModel` and will end up internally as `testModel` with our getter function. This will manifest a model name for future versions which is not aligned to Ember-Datas naming schema. I think, it is ok to use your own normalization, but it is a decision affecting the future.

On the other hand, how is Ember-Data handling the incompatibility?

fsmanuel:
@OleRoel here is the blog post describing the deprecation:
http://emberjs.com/blog/2015/05/21/ember-data-1-0-beta-18-released.html

OleRoel:
@fsmanuel thanks, I've already read the blog post. But, why not ask there :-)

http://emberjs.com/blog/2015/05/21/ember-data-1-0-beta-18-released.html#comment-2066826791

broerse:
Can we `if` it like this: https://github.com/halfdan/emberfire/commit/e41e1166db104065a430970480758ecea4d2630c

rsutphin:
I think I might not have done a good job describing my concern. Say you have this model:

```
// app/models/taco-soup.js
import DS from 'ember-data';

export DS.Model.extend({
  rev: DS.attr('string');
  flavor: DS.attr('string');
});
```

Your application uses the current version of ember-pouch and you do:

```
var alPastor = store.createRecord('taco-soup', { id: 'TS1', flavor: 'Al pastor' });
alPastor.save();
```

Then you upgrade to a theoretical future version with this PR applied. You do:

```
store.find('taco-soup', 'TS1')
```

**This will not find the previously created record.**

(Freehanded code; apologies for any errors.)

What you call your model (i.e., whether you pass `'taco-soup'` or `'tacoSoup'` when you call the store) does not matter — your app will no longer be able to read previously written records. There's nothing you can do in your app alone (i.e., without library patching or data changes) to fix this.

This is because ember-pouch (via relational-pouch) used the `typeKey` to generate the document IDs for your documents. If you change from passing the camelized `typeKey` to the dasherized `modelName` into relational-pouch, all your old documents will effectively disappear.

---

@OleRoel, your proposed function is what I had in mind. To be clear, though, I would not call it `getModelName` in order to prevent the sort of conceptual mismatch with ED that you're talking about. I'd call it something like `getRecordTypeName` or `getTypeNameForDocumentId` to indicate its purpose.

With that change I wouldn't really worry about this being different from ember-data — the fact that the document IDs incorporate the model name / type key is an internal implementation detail of ember-pouch. Users who only interact with ember-pouch via the ember-data store will never be aware of this.

---

> On the other hand, how is Ember-Data handling the incompatibility?

@OleRoel, for ember-data's main use case (retrieving data from a remote server), the incompatibility can be handled in your custom adapter/serializer, as outlined in the blog post. The solution I'm proposing for ember-pouch is roughly equivalent to that — it's like if your server had a resource named `/tacoSoup/:id` that you used to get/put taco soup records. If previously you were using doing something like

```
`${type.typeKey}/${id}`
```

to construct your URLs, that blog post advises you to do

```
`${camelize(type.modelName)}/${id}`
```

instead.

> Can we if it like this: halfdan/emberfire@e41e116

@broerse, that solution only preserves syntactic backwards compatibility. I don't know much about Firebase, but unless it internally considers `taco-soup` and `tacoSoup` to be equivalent, I would guess that that patch would have the same data-losing problem that I'm talking about here.

rsutphin:
Sorry about the wall of text :smile:, in which I failed to mention that I'm planning to submit a PR for my suggestion tonight.

broerse:
@rsutphin Looking forward to it. :+1:

OleRoel:
@rsutphin :+1:

nolanlawson:
fixed by https://github.com/nolanlawson/ember-pouch/pull/64, thanks for the PR to start the conversation, though!

Issue #62, opened 2015-06-02
Usage of `typeKey` has been deprecated warning with Ember-Data 1.0.0-beta.18


OleRoel:
I get a lot of 'DEPRECATION: Usage of `typeKey` has been deprecated and will be removed in Ember Data 1.0. It has been replaced by `modelName` on the model class.' messages in my console lately.

The warning has been introduced with 1.0.0-beta.18:

https://github.com/emberjs/data/blob/master/CHANGELOG.md#release-100-beta18-may-18-2015

OleRoel:
Not sure if the fix is really that simple, but a quick find/replace of `typeKey` with `modelName` in addon/adapters/pouch.js makes the deprecation warning go away. Time for a PR? And what's about backward compatibility?

nolanlawson:
fixed by https://github.com/nolanlawson/ember-pouch/pull/64

PR #64, opened 2015-06-08
Use modelName instead of typeKey when available, but preserve existing ID generation behavior

rsutphin:
This is my suggestion for #62, based on my desire to prevent all existing ember-pouch users from having to do a data migration as discussed in #63.

It also includes a set of very basic integration tests. If this feature doesn't get integrated, they should still be usable on their own. (I wrote them to verify that this change did preserve the prior behavior.)

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -12,6 +12,7 @@ const {
   on,
   String: {
     pluralize,
+    camelize,
     classify
   }
 } = Ember;
@@ -72,21 +73,22 @@ export default DS.RESTAdapter.extend({
   },
 
   _init: function (type) {
-    var self = this;
+    var self = this,
+        recordTypeName = this.getRecordTypeName(type);
     if (!this.db || typeof this.db !== 'object') {
       throw new Error('Please set the `db` property on the adapter.');
     }
 
     if (!Ember.get(type, 'attributes').has('rev')) {
-      var modelName = classify(type.typeKey);
+      var modelName = classify(recordTypeName);
       throw new Error('Please add a `rev` attribute of type `string`' +
         ' on the ' + modelName + ' model.');
     }
 
     this._schema = this._schema || [];
 
-    var singular = type.typeKey;
-    var plural = pluralize(type.typeKey);
+    var singular = recordTypeName;
+    var plural = pluralize(recordTypeName);
 
     // check that we haven't already registered this model
     for (var i = 0, len = this._schema.length; i < len; i++) {
@@ -131,7 +133,8 @@ export default DS.RESTAdapter.extend({
 
   _recordToData: function (store, type, record) {
     var data = {};
-    var serializer = store.serializerFor(type.typeKey);
+    var recordTypeName = this.getRecordTypeName(type);
+    var serializer = store.serializerFor(recordTypeName);
 
     var recordToStore = record;
     // In Ember-Data beta.15, we need to take a snapshot. See issue #45.
@@ -149,7 +152,7 @@ export default DS.RESTAdapter.extend({
       {includeId: true}
     );
 
-    data = data[type.typeKey];
+    data = data[recordTypeName];
 
     // ember sets it to null automatically. don't need it.
     if (data.rev === null) {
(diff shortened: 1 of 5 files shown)


rsutphin:
The integration tests work with PhantomJS 2.0 only, not 1.9. I thought from reading the `.travis.yml` that PhantomJS 2.0 was available on travis, but it appears not yet (travis-ci/travis-ci#3225).

So, the CI build is going to fail. I'll look into getting PhantomJS 2.0 available for the build.

broerse:
:+1:

OleRoel:
:+1:  Looks good to me

broerse:
Tested on https://github.com/broerse/ember-cli-blog and it works fine!

fsmanuel:
@rsutphin looks good but can we use the new terminology and do
`recordModelName = this.getRecordModelName(type);`

rsutphin:
@fsmanuel, I deliberately used a different term from either `modelName` or `typeKey` in order to prevent the sort of conceptual confusion @OleRoel and I discussed on #63:

> I would not call it `getModelName` in order to prevent the sort of conceptual mismatch with ED that you're talking about. I'd call it something like `getRecordTypeName` or `getTypeNameForDocumentId` to indicate its purpose.

Do you think _not_ calling it something that includes `modelName` is more confusing?

rsutphin:
Squashed the various CI-fixing commits.

addon/adapters/pouch.js
```diff
@@ -131,7 +133,8 @@ export default DS.RESTAdapter.extend({
 
   _recordToData: function (store, type, record) {
     var data = {};
-    var serializer = store.serializerFor(type.typeKey);
+    var recordTypeName = this.getRecordTypeName(type);
+    var serializer = store.serializerFor(recordTypeName);
```
fsmanuel: I think we get a problem here when ED transitions to dasherized types.

rsutphin: That's a good point. I initially used `type.modelName || type.typeKey` instead of `getRecordTypeName` in this method for that reason. But the serialized data comes back with a camelized key (see line 155 below), and I figured it wouldn't hurt to use the camelized version in both places.

ED has supported either dasherized or camelized names for lookups (`modelFor`, `serializerFor`, etc) for a while — have you seen something that indicates that it is going to stop supporting camelized ones at some point?

rsutphin: Thinking about this some more — these uses are conceptually different from `recordTypeName`, so it's probably best not to use it here. E.g., if you override `getRecordTypeName` to provide your own scheme, the serializer is still going to return the data with a camelized key. And if the serializer changes in the future, it shouldn't affect the type name used in the record IDs. So I'll change this.

fsmanuel:
@rsutphin i see your point. What do you think about two functions:

```js
modelNameFromType(type) {
  if (type.modelName) {
    return type.modelName;
  } else {
    // This branch can be removed when the library drops support for
    // ember-data 1.0-beta17 and earlier.
    return type.typeKey;
  }
}
```

```js
recordTypeName(type) {
  if (this.get('useLegacyTypes')) {
    return camelize(this.modelNameFromType(type));
  }
  
  return this.modelNameFromType(type);
}
```

If we call it a breaking change new users can use the dasherized types and old users should set `useLegacyTypes`. That way we can differentiate between ED and pouch internals (see my comment on `store.serializerFor` above).

rsutphin:
@fsmanuel, we could do that, but it seems unnecessary. Can you help me understand why you think this complexity (including a new option to document and support, communicating the breaking change to all existing users, etc.) is worth it just to use dasherized names in document IDs instead of camelized ones? The specific details of the document IDs aren't exposed to regular users of the adapter, so the risk of confusion with what ember-data does is very low.

I agree that the interaction with the serializer needs to be addressed; I'll push something up shortly to handle that.

addon/adapters/pouch.js
rsutphin: Addressed in 4b33181.

fsmanuel:
@rsutphin :+1:

nolanlawson:
Nice work! Love the cross-version solution and the new tests. And nice hack to get PhantomJS 2.0 working; I didn't realize it was possible to run it in Travis. :)

Thanks for all the input, everyone! I'll publish a patch version today.

nolanlawson:
2.0.1 published and changelog in the README updated. Thanks again everyone!

PR #65, opened 2015-06-11
Missed one use of typeKey in #64.

rsutphin:
Not ready to merge, because the new test does not pass on ember-data 1.0-beta19 or 19.1. The problem appears to be that the relationship returned by eachRelationship has the name of the associated type as `type` instead of the DS.Model object. It works fine on older versions (tested beta17 and 18).

Have not tried in a real app yet to see if this is a test artifact or an ember-data bug.


broerse:
:+1:

Issue #66, opened 2015-06-17
Relationship in model generates "Cannot read property ‘has’ of undefined' exception" with Ember Data 1.13.0


OleRoel:
When I use a relationship in my model like:

```javascript
import DS from "ember-data";
import { Model } from 'ember-pouch';

var Modelname = Model.extend({
	name: DS.attr('string', {defaultValue: ""}),
	pages: DS.hasMany('page', {async: true, serialize: true})
});

export default Modelname;
```

The code is borrowed from @broerse from [where the discussion started](http://discuss.emberjs.com/t/cannot-read-property-has-of-undefined/8187). The problem seems to be introduced with verson 19.1 of ember-data as brianmbutler states in the original discussion.

The same exception pops up with `DS.belongsTo`.

With help of the Chrome debugger I was able to find the culprit in the init method.

```javascript
...
if (!Ember.get(type, 'attributes').has('rev')) {
...
```

https://github.com/nolanlawson/ember-pouch/blob/master/addon/adapters/pouch.js#L82

expects type to be a model object but gets a string when a relationship is defined inside your model.

OleRoel:
The `store.findAll` seems to trigger the exception. This ends up in the corresponding adapter method

https://github.com/nolanlawson/ember-pouch/blob/master/addon/adapters/pouch.js#L191

where `this._init(type)`gets called with a string.

rsutphin:
This is the problem that's causing the tests for #65 to fail, also. It's a change between 1.0-beta18 and 1.0-beta19. I think it's unintended because of the documentation you cite, @OleRoel. I've been continuing to use beta18 until I have a chance to work up a reproducible case for ED.

OleRoel:
Just reported what appears to me an ember-data bug [here](https://github.com/emberjs/data/issues/3373).

OleRoel:
Just created an PR for this issue.

@rsutphin Can you give me a little guidance of how to run the tests? The integration-tests are failing here with 'Can't find variable: PouchDB' errors when I simply run `ember test`on the command line.

rsutphin:
@OleRoel, did you run `npm install` and `bower install` before you tried to run the tests?

OleRoel:
@rsutphin Yes, I did. I'll try to find out what I am doing wrong tonight.

rsutphin:
[...]

PR #67, opened 2015-06-18
PR for issue #66

OleRoel:
This fixes a breaking change in ember-data beta 19, see https://github.com/emberjs/data/issues/3373.
I've expanded the _init method with an extra `store` argument, allowing lookup of model by name inside the method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,11 @@ cache:
     - node_modules
 
 env:
-  - EMBER_TRY_SCENARIO=default
+  - EMBER_TRY_SCENARIO=ember-data-beta.15
+  - EMBER_TRY_SCENARIO=ember-data-beta.16
+  - EMBER_TRY_SCENARIO=ember-data-beta.18
+  - EMBER_TRY_SCENARIO=ember-data-beta.19
+  - EMBER_TRY_SCENARIO=ember-data-1.13
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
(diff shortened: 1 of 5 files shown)


rsutphin:
This PR looks like it was not made over current master. Can you rebase it to make your changes clearer?

rsutphin:
The thing that's causing the ember-release travis build to fail is fixed in #68. The ember-beta CI failure still happens with #68 applied, but I don't think it should block this change.

nolanlawson:
I merged #68, so @OleRoel could you please rebase this PR on master so Travis can verify that it's green?

nolanlawson:
Also I suppose we will have to update the ember-try configuration to demonstrate that this fixes ember-data beta 19?

rsutphin:
There aren't currently tests that break for #66, so the ember-try configuration for beta19 is passing. The test from #65 would show this bug and could be ported out to this PR.

OleRoel:
Big sight... I had some fun with git rebase and was unsure that I did everything correctly. But it seems as things worked out.

MKruschke:
hi, i tested the 66 branch against my application but found another issue. There is missing a second parameter for binding by calling the eachrelationship function.  

My application runs under ember 1.13.1 and ember-data 1.13.1

Here my patch to fix that.

```
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -131,7 +131,7 @@ export default DS.RESTAdapter.extend({
         schemaDef.relations[rel.key] = relDef;
         self._init(store, relModel);
       }
-    });
+    },this);

     this.db.setSchema(this._schema);
   },

```

OleRoel:
Yes, that was my mistake by doing a quick code improvement without really looking into the code in response to another comment. I promise, I will never show this bad behavior again!

@brianmbutler already found the bug and commented about it:

https://github.com/nolanlawson/ember-pouch/issues/66#issuecomment-113346734

@rsutphin also mentioned to fix this in the PR https://github.com/nolanlawson/ember-pouch/pull/69
https://github.com/nolanlawson/ember-pouch/pull/69#commitcomment-11774632

rsutphin:
Merged in #69.

PR #68, opened 2015-06-18
Ember-try config for all supported versions of ember-data, plus fixes for the tests on ember-data beta19 and later

rsutphin:
There is code in place to support versions back to beta.15, so it seems like we should have the tests run that far back. We can always drop support in the future.

Each ember-data version is paired with the latest bugfix of the ember version that was current when that ember-data version was released.

This PR also makes it so that a failure in one of the symbolic versions (release, beta, canary) does not make the whole build a failure. Rationale: If there’s an incompatibility between a future version of ember or ember-data and ember-pouch, that should be fixed. But it shouldn’t block accepting changes that do pass on all the supported versions. (If there were a way to have travis run a separate build whenever there was a new release/beta/canary version, that would be great. But I don't think there is.)

This PR also fixes an incompatibility between the existing integration tests and ember-data beta19.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,11 @@ cache:
     - node_modules
 
 env:
-  - EMBER_TRY_SCENARIO=default
+  - EMBER_TRY_SCENARIO=ember-data-beta.15
+  - EMBER_TRY_SCENARIO=ember-data-beta.16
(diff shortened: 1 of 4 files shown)


nolanlawson:
Fantastic work, LGTM. I agree that being explicit about which versions we support is best. Maybe we should also update the README?

PR #69, opened 2015-06-19
Fix #66, breaking change in ember-data beta 19

### LINKED ISSUE #66: Relationship in model generates "Cannot read property ‘has’ of undefined' exception" with Ember Data 1.13.0

nolanlawson:
I'm still not entirely clear what's going on in https://github.com/nolanlawson/ember-pouch/pull/67, but @OleRoel I merged your commits into a single, squashed commit. @rsutphin does anything need to change in the travis config to make this green? Should we add a new test?

(And yes, even though this is my repo, I am pretty overwhelmed by all the changes in Ember Data recently and having a hard time keeping up. :stuck_out_tongue:)

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -72,7 +72,7 @@ export default DS.RESTAdapter.extend({
     }
   },
 
-  _init: function (type) {
+  _init: function (store, type) {
     var self = this,
         recordTypeName = this.getRecordTypeName(type);
     if (!this.db || typeof this.db !== 'object') {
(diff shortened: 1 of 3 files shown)


rsutphin:
We should add a new test. The test in #65 should cover this case. I can add it here, but I won't be able to get to it until later next week.

OleRoel:
> One of the comments on #66 pointed out that this line will fail because this is undefined here. The eachRelationship callback function needs a bind(this) at the end, or it can use self instead of this.

I was on the wrong track and added an extra `if`statement for non extsitent models and ignored the context of the code. Mea culpa! But this means, we need tests for related models.

@nolanlawson I am sorry when I created extra work for you with my rebase attempt.

OleRoel:
Are we going to get a version number bump, too? (Sorry for being impatient!)

PR #70, opened 2015-06-21
update README.md

brianmbutler:
Remove reference to typkey as it was deprecated, unless you want to leave for reference for older versions of ember-data.


rsutphin:
Thanks! Can you rebase to get rid of that extraneous merge commit?

brianmbutler:
Not sure I know how to do that, I made the change right in the text editor on github.com <http://github.com/>

Issue #72, opened 2015-06-27
Adapter#find has been deprecated and renamed to `findRecord`.


OleRoel:
This has been introduced with ED 1.13, the PR can be found here: https://github.com/emberjs/data/pull/3306. We should keep a proxy ` find` method for backwards compatibilty which just forwards to  `findRecord`. ~~It's too late here for me to set up a PR, anyone who likes to jump in?~~

OleRoel:
There are more deprecation warnings with ember-pouch:
  * Using store.getById() has been deprecated. Use store.peekRecord to get a record by a given type and ID without triggering a fetch
  * DS.Model#isDirty has been deprecated please use hasDirtyAttributes instead
  * You tried to look up 'store:main', but this has been deprecated in favor of 'service:store'

Not sure how to handle this one (just ignore it?):
  * The default behavior of shouldReloadAll will change in Ember Data 2.0 to always return false when there is at least one "agent" record in the store. If you would like to preserve the current behavior please override shouldReloadAll in you adapter:application and return true.

OleRoel:
A table with ED store API changes can be found here: http://emberjs.com/blog/2015/06/18/ember-data-1-13-released.html

Ember beta.19| Ember 1.13
------------ | -------------
`store.getById(type, id)` | `store.peekRecord(type, id)`
`store.all(type)`| `store.peekAll(type)`
`store.find(type, id)`| `store.findRecord(type, id, options)`
`store.find(type)`  | `store.findAll(type, options)`
N/A | `store.queryRecord(type, query)`
`store.find(type, { query })` | `store.query(type, query)`
`store.fetch(type, id)` | `store.findRecord(type, id, {reload: true})`

rsutphin:
> There are more deprecation warnings with ember-pouch:

I opened a new issue for these — #78.

> Not sure how to handle this one (just ignore it?):

I think this is one that needs to be decided on for each application.

I'm going to close this issue now since the main topic was fixed by #75.

PR #73, opened 2015-06-28
PR for issue #72

OleRoel:
Renamed `find`to `findRecord` and added `find`method, which forward calls to `findRecord`to keep backward compatibility.

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -72,7 +72,7 @@ export default DS.RESTAdapter.extend({
     }
   },
 
-  _init: function (type) {
+  _init: function (store, type) {
     var self = this,
         recordTypeName = this.getRecordTypeName(type);
     if (!this.db || typeof this.db !== 'object') {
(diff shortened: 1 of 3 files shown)


rsutphin:
@OleRoel, thanks for taking this on.

It looks like you've got some extra commits in your PR branch, like there were in #67. We'd want to avoid this to keep the history in the main repo clear and easy to understand after the PR is merged.

To avoid getting into this situation, you want to do two things:

* Before you make your feature branch, make sure your local master & and your fork's master branch match the current master here, in upstream.
* Don't ever merge into the feature branch. If you need to bring in upstream changes before you submit your PR, use rebase instead.

[This is a detailed overview of how you can and why you would do this](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request). Here's a summary for your current situation, but be careful because some of these actions are destructive if you do them wrong. These instructions assume that your fork is checked out and its remote is called `origin` (if you `git clone`d your fork, this will be the case.)

* Make a feature branch from your current master (since it has the changes you want to submit on it).

```
git branch 72-deprecated_find
```

* Retrieve the current upstream.

```
git remote add upstream git@github.com:nolanlawson/ember-pouch.git
git fetch --all
```

* Reset your local master to match upstream (this is a destructive step).

```bash
# ensure you are on master
git checkout master
git reset --hard upstream/master
```

* Update your remote master to match your new local, which matches upstream (this is also destructive). This is optional, but it will probably reduce confusion down the line.

```
git push --force origin master
```

* Rebase your feature branch over the new local master

```
git checkout 72-deprecated_find
git rebase master
```

There may be conflicts at this point. Follow the rebase prompts to resolve them. Once you are sure that the rebase was successful and your changes are present in the feature branch, continue.

* Close this PR, since it points at your master branch.
* Push the new feature branch to your fork

```
# Make sure you are on the feature branch
git checkout 72-deprecated_find
git push -u origin 72-deprecated_find
```

* Make a new PR based on the new feature branch.

OleRoel:
@rsutphin Thank you for your detailed instructions, they are really appreciated. I am having a hard time to figure out how to use rebase and was thinking (hoping) to do it right this time, obvisouly I didn't get it again. The stupid thing is, that I am trying things out in public. So, please accept my excuse again, I don't want to create extra work here.

OleRoel:
@rsutphin ...and I am now a bit fearful to push the hard reset master to my Git-repo, because I don't know what happens to the PR then. Should I execute another experiment in public and everyone promises to close their eyes while I am doing it? Or do you like to create a clean PR with the necessary changes and I will do my rebase afterwards? I don't mind doing a stunt while no one is watching.

rsutphin:
I think the worst thing that would happen is that the PR would get automatically closed (because it has no changes in in any more). But I don't think anything terrible/unrecoverable would happen. I think it will be an interesting experiment and no one will hold anything against you regardless.

But if you really don't want to do it, you can always skip the step about force pushing your new local master. You can continue with the rest of the steps without doing that one.

nolanlawson:
Thanks a lot; rebasing is definitely the preferred method since the Git repo doesn't end up looking like a subway map! :)

For the record @rsutphin, if there are no merge conflicts, I will often just manually rebase and squash the commits myself, and then change the author name to preserve original authorship. I find the whole rebasing thing to be a really difficult and error-prone thing, especially for newcomers. :P

PR #74, opened 2015-06-28
Implement queries #7

leojpod:
This is a temporary implementation to support the queries in ember-pouch

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -204,11 +204,15 @@ export default DS.RESTAdapter.extend({
     return this.db.rel.find(this.getRecordTypeName(type), ids);
   },
 
-  testQuery: function (record, query) {
-    for(let item in query) {
-      if (query.hasOwnProperty(item)) {
-        const value = query[item];
-        if (record.get(item) !== value) {
+  matchQuery: function (record, query) {
+    for(let property in query) {
+      if (query.hasOwnProperty(property)) {
+        const value = query[property];
+        if (Object.prototype.toString.call(value) === '[object RegExp]') {
+          if (!value.test(record[property])) {
+            return false;
+          }
+        } else if (record[property] !== value) {
           return false;
         }
       }
@@ -219,13 +223,15 @@ export default DS.RESTAdapter.extend({
 
   findQuery: function(store, type, query) {
     var self = this;
-    this._init(type);
+    this._init(store, type);
     var recordTypeName = this.getRecordTypeName(type);
     return this.db.rel.find(recordTypeName).then((res) => {
-      Ember.Logger.debug('findQuery -> found ', res.get('length'), ' records (', type, ')');
-      return res.filter((record) => {
-        self.testQuery(record, query);
+      const records = res[pluralize(recordTypeName)];
+      const results = {};
+      results[pluralize(recordTypeName)] = records.filter((record) => {
+        return self.matchQuery(record, query);
       });
+      return results;
     });
   },
 
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -204,10 +204,29 @@ export default DS.RESTAdapter.extend({
     return this.db.rel.find(this.getRecordTypeName(type), ids);
   },
 
-  findQuery: function(/* store, type, query */) {
-    throw new Error(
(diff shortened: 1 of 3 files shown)


leojpod:
@nolanlawson This works but I do not like it. I wish to work with `pouchdb-find` but I've some issue with that. I guess `pouchdb-find` should be integrated with `relational-pouch` but I did not find the right way to do it.

broerse:
@leojpod Interesting. Is it not better in this case to use https://github.com/brianreavis/sifter.js/ than write your own?

leojpod:
@broerse not sure if it is worse the trouble for his case: it is a temporary solution, using `pouchdb-find` would make more sense and would give better performance I guess and I am not sure which other query option should be added. 

Do you think we should add support for more than regex and straight equal ?

broerse:
Regex and straight equal is fine. I only tend to us sifter.js when I need filtering and was wondering if I would use this. Thanks for this PR.

nolanlawson:
I do think that the first step ought to be to integrate pouchdb-find into relational-pouch. Sorry for the radio silence, but I've been on vacation for the past few weeks. :)

This next weekend I'll try to do a big push to get pouchdb-find closer to production status, and to get it integrated into relational-pouch. I know it's been a big ask from the community, so it's number one on my priority list.

nolanlawson:
Related to https://github.com/nolanlawson/relational-pouch/issues/3

patuku-roy:
@nolanlawson just asking about what you are going to work [on it](https://github.com/nolanlawson/ember-pouch/pull/74#issuecomment-121955298) , is it related to [hasMany issue](https://github.com/nolanlawson/ember-pouch/issues/16) ? because I believe this is something need to solve soon because those issues has very close relation. Btw, nice work ... thanks a lot

broerse:
@patuku-roy The hasMany issue should be fixed with `serialize:true takes priority over the OneToMany check for relationships`. See https://github.com/emberjs/data/pull/3214. I started testing but I did not have more time to check it out. Please report back if it works.

fsmanuel:
I'm not sure when ember-data added `_shouldSerializeHasMany` but in 1.13.x it is available
We can add that line to the serializer and add a section to the readme that ppl should import the serializer.

```js
  _shouldSerializeHasMany: function() { return true; }
```

patuku-roy:
@broerse I don't have a chance yet to try that way, I will post the feedback if I have a chance, meanwhile I'm working on other stuff, while waiting the query stuff in here.

@fsmanuel unfortunately I'm stuck with ED Beta-18 right now, the project has been too complicated to port to latest ED, do you have a chance looking at ED Beta-18 yet?

@nolanlawson OOT, since the npm maintainer has been changed, are you still working on this? I really keen to hear the update about the query feature, this is a simple thing but very crucial on my project :) ... I've been tried integrate the pouchdb-find to ember-pouch but I guess I did the wrong direction because pouchdb-find suppose to be integrate inside relational-pouch rather than directly inside ember-pouch. So any updates about this, will be really helpful ... and really thanks for your great work :+1: :+1: :+1:

nolanlawson:
Yes, I have a bunch of changes in store for pouchdb-find that I hope to integrate into relational-pouch (one PouchDB plugin can depend on another; it's easy). pouchdb-find improvements will be part of my project for the summer. :)

patuku-roy:
@nolanlawson Cool, can't wait ... Thanks in advance for your work :) ... :+1: :+1: :+1:

nolanlawson:
Sorry, my bad for leaving this PR hanging for so long. The tests look fantastic, and I would love to see this re-worked into a new PR, but I do think the right approach is to use `pouchdb-find`, which maybe should be integrated into `relational-pouch` first. (Or maybe not! the `relational-pouch` API does expose ways to translate between the relation-y IDs and the normal PouchDB `_ids`, so it's possible that you could integrate both into `ember-pouch` without changing `relational-pouch` at all.)

In any case, I'm gonna close for now, please rebase and reopen if you have a chance to work on it.

leojpod:
@nolanlawson No problem, this was just a temp fix :) I agree that it should be done in another way. 

I had looked at the `relational-pouch` code but I wasn't really sure of where to plug the `pouchdb-find`. I will look at it again another day (if no one does it before!)

Issue #79, opened 2015-07-19
shouldReloadAll will change in Ember Data 2.0


broerse:
I get this deprecation warning:
`Ember Inspector (Deprecation Trace): The default behavior of shouldReloadAll will change in Ember Data 2.0 to always return false when there is at least one "author" record in the store. If you would like to preserve the current behavior please override shouldReloadAll in your adapter:application and return true.`
Do we need to change something? Is the new behavior better?

fsmanuel:
The new behavior aims to optimize performance.
There are 4 methods:
- [shouldReloadRecord](http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_shouldReloadRecord)
- [shouldBackgroundReloadRecord](http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_shouldBackgroundReloadRecord)
- [shouldReloadAll](http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_shouldReloadAll)
- [shouldBackgroundReloadAll](http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_shouldBackgroundReloadAll)

The idea in Ember Data 2.0 is to return a cached version immediately and than refetch in the background.

For pouchdb the fetch isn't expensive because we don't need an ajax request. I guess it's best to always reload:

```js
  // Reload behavior
  shouldReloadRecord: function() { return true; },
  shouldReloadAll: function() { return true; },
  shouldBackgroundReloadRecord: function() { return true; },
  shouldBackgroundReloadAll: function() { return true; }
```

broerse:
Perhaps we should implement this in the ember-pouch adapter by default.

fsmanuel:
:+1:

broerse:
I will write a PR for this.

rsutphin:
The change watcher already provides automatic refreshing of cached records in a way that's more efficient than this ember-data feature — it updates the store-cached record as soon as it is changed (not just when `findRecord` is called), and it only updates when it is actually different (instead of making a query every time `findRecord` is called). So I think returning true from `shouldReloadRecord` or `shouldBackgroundReloadRecord` in an ember-pouch app will make the app do more work without benefit. I'd be in favor of having ember-pouch default both of those to `false`.

[...]

PR #80, opened 2015-07-19
Add Ember Data 2.0 Reload behavior

broerse:
This fixes the https://github.com/nolanlawson/ember-pouch/issues/79 deprecation warning and preserves the current behavior until we know the new Reload behavior works and is faster in Ember Data 2.0

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -20,6 +20,12 @@ const {
 export default DS.RESTAdapter.extend({
   coalesceFindRequests: true,
 
+  // Ember Data 2.0 Reload behavior
+  shouldReloadRecord: function() { return true; },
+  shouldReloadAll: function() { return true; },
+  shouldBackgroundReloadRecord: function() { return true; },
+  shouldBackgroundReloadAll: function() { return true; },
+
   _startChangesToStoreListener: on('init', function () {
     this.changes = this.db.changes({
       since: 'now',


fsmanuel:
:+1:

rsutphin:
@broerse, thanks for taking the time to submit this.

There's more detail in my comment on #79, but briefy: I don't think this is a good idea. If I understand things correctly, `shouldReloadRecord` and `shouldBackgroundReloadRecord` have no benefits over our existing change watcher. They will make your app do extra work for no gain. If anything, we should default them to `false`.

I'm less sure about `shouldReloadAll` and `shouldBackgroundReloadAll`. There are pluses and minuses to either approach, so I feel like inheriting the ember-data behavior (whatever it may be) is the best thing to so. The adapter will behave the way the default adapters are documented to behave. If folks want to change that behavior, they will have that option.

broerse:
@rsutphin @fsmanuel  I don't think I fully agree but I am fine with closing this.

fsmanuel:
What about adding `shouldReloadRecord: function() { return false; }` to remove the deprecation warning. I haven't checked ember-data code to verify that the deprecation will actually go away...

broerse:
@fsmanuel If that removes the deprecation warning (not checked yet) we shout do all 4 because I get `shouldReloadAll` deprecation warnings in my apps.

rsutphin:
> we shout do all 4 because I get `shouldReloadAll` deprecation warnings in my apps. 

We should only do all four if we think there's a reason why ember-pouch should specify a value for each one. If not, it should be left up to the person using ember-pouch in an application. 

It's not the responsibility of this library to get rid of all deprecation warnings — if you were using a built-in adapter (e.g., `RESTAdapter`), you'd be seeing the same deprecation warning. It's communicating an upcoming behavior change that you, as an application developer, need to be aware of. It's not the same as the warning in #72 (e.g.) which communicates changes to the structure of the adapter API and so is the responsibility of the library.

rsutphin:
> What about adding `shouldReloadRecord: function() { return false; }` to remove the deprecation warning.

Deprecation warnings are emitted for `shouldReloadAll` and `shouldBackgroundReloadRecord` only:

https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/adapter.js#L452-L529

I would be fine with having ember-pouch override `shouldBackgroundReloadRecord` to be `false`, since background reloading of individual records is not necessary in ember-pouch. (`shouldReloadRecord` is already `false` and apparently will continue to be `false` in 2.0.)

For the reasons I've outlined above and elsewhere, I don't think that ember-pouch should make a decision about `shouldReloadAll`. It should leave the warning in place and let individual application developers decide what to do.

fsmanuel:
Sounds reasonable!

broerse:
Just checked. Return false also disables the deprecation warnings but it does not preserve the current behavior. I still think that until we know how the new Reload behavior works we should merge this. It gives us some time to test things out. If I am on my own here I am fine with closing this.

rsutphin:
> Return false also disables the deprecation warnings but it does not preserve the current behavior.

@broerse, what I'm hearing from you is that you want to preserve the existing ember 1.x behavior. Regardless of whether we agree about that, this pull request does not preserve the existing behavior. The existing behavior is what `DS.Adapter` in 1.13 does:

* `shouldReloadRecord` => `false`
* `shouldReloadAll` => `true`

I'm going to close this because I think the consensus is that we will override the single-record methods to return false and continue to do so in ember-data 2.0 — the change watcher means that ember-pouch never needs single records reloaded. We will leave the default behavior for record arrays because there's not a clear correct behavior based on the adapter alone — it needs to be decided by each application developer and so keeping the ember-data default behavior is the least confusing path.

If you disagree with either of those points, please comment.

Issue #81, opened 2015-07-20
Check the rev to avoid unnecessary reloads in change watcher


rsutphin:
The change watcher calls `reload` when a change is encountered for a record that is in ember-data's store cache. However, the change notifier in PouchDB notifies for all changes, including ones that originated in the ember app. Compare the rev of the change to the rev the store knows about and avoid reloading for those changes.

Originally suggested by @broerse in comments on #50.

broerse:
My changes seem to work fine but I get some test errors. Please take a look.

rsutphin:
While it's true that the change notifier fires for local changes, it doesn't actually call `reload`. At the point that the change watcher is called, the ember-data record is still dirty. The change watcher doesn't reload dirty records to avoid losing local changes. 

So, there's no point in adding a separate rev check. Apologies, I should have looked into this more thoroughly before opening this ticket.

Issue #50, opened 2015-03-25
ember-pouch live updating stopped working


broerse:
On: https://github.com/broerse/ember-cli-blog/tree/master/app/adapters the live updating stopped working. If you now change something in one browser you don't see the change until you switch to an other route and back. I tested with ember-pouch 1.2.3 and 1.2.4 and ember-data beta15 & beta16 but snapshots don't seem to be the problem.

broerse:
Update:
- When a new record is created in browser 1 it is synced to the remote CouchDB server. It is not synced back to Ember-data from browser 2. (You can see this in the ember inspector) It is synced back to the IndexDB from browser 2.
- When you then create a record in browser 2 this is also synced to the remote CouchDB server but the record from step 1 does not show up in Ember-Data but shows up in the IndexDB.

So it seems syncing between Pouch and Ember-Data is not working 100% anymore.

rsutphin:
@broerse, I think you might be seeing a side-effect of #43. That PR changed the default change watcher so that it only updates records that are already loaded into ember-data's store (i.e., with a find). It no longer loads every record it sees into the store. (This is necessary to support replicating with large databases where you don't need all the data available to Ember immediately.)

If you do want to have every record (or every record of a particular type) automatically loaded as it comes in, I suggest adding a change watcher that's specific to your app's needs.

broerse:
@rsutphin Perhaps but it breaks how it worked before. Perhaps we need a default change watcher that     does work like it was before and extend if we want something else. (Just like Ember ;-)) I understand why this is basically a good idea. 
I think most people starting with ember-pouch want sync. like firebase and don't have large databases.  Can you show me how to add the "old" behavior change watcher? 
@nolanlawson do we need a major version update and some doc's for this break?

nolanlawson:
My bad, I didn't realize it was a breaking change. It's already published, so that ship has kind of sailed :sweat:.

Looking at #43, I don't really understand how it introduced a breaking change. In both case, we are doing `changes({since: 'now'})`, which means that it shouldn't trigger for pre-existing documents. @rsutphin I also don't quite follow your explanation.

In general I think the userbase for ember-pouch is low enough that we can feel comfortable moving fast and breaking things. Although if you think it's serious, I can unpublish from npm.

@broerse can you rearchitect your app to use the new system, or do you absolutely positively need the old functionality?

broerse:
On my phone. No problem breaking things. I can comment tomorrow.

[...]

PR #82, opened 2015-07-23
Check the rev to avoid unnecessary reloads in change watcher

broerse:
PR for https://github.com/nolanlawson/ember-pouch/issues/81

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -37,7 +37,7 @@ export default DS.RESTAdapter.extend({
     // skip changes for non-relational_pouch docs. E.g., design docs.
     if (!obj.type || !obj.id || obj.type === '') { return; }
 
-    var store = this.container.lookup('store:main');
+    var store = this.container.lookup('service:store');
 
     try {
       store.modelFor(obj.type);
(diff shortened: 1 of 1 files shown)


rsutphin:
The tests are mostly failing because of the changes that remove backwards compatibility with older ember-datas. The 1.13 tests are failing because the asynchronous reload introduced in the case where the record is still loading when the change occurs assumes that the record will be loaded in the next tick. That's not guaranteed.

I don't think any of this is necessary. Checking if the rev is different is simply a matter of adding another condition to this `if`: https://github.com/nolanlawson/ember-pouch/blob/master/addon/adapters/pouch.js#L55

broerse:
I don know how to do the `if` without adding asynchronicity. https://github.com/nolanlawson/ember-pouch/blob/master/addon/adapters/pouch.js#L51 is true the next ember run if you make the change. It is false if the change was made remote. The same it true for https://github.com/nolanlawson/ember-pouch/blob/master/addon/adapters/pouch.js#L55. The rev in Ember Data is updated the next ember run if you make the change.

If it is simply a matter of adding another condition to this `if` how is it done?

https://github.com/nolanlawson/ember-pouch/issues/50 is the issue that needs to be solved. I believe this solves it.

rsutphin:
(I changed the URLs to match the line numbers when you made your comment @broerse — the lines have shifted since I merged #83.)

> https://github.com/nolanlawson/ember-pouch/blob/c15b0850d52512fde3ed5860e071f50da9c6d4db/addon/adapters/pouch.js#L51 is true the next ember run if you make the change. It is false if the change was made remote.

It's also true immediately if you made the change. In either case it may or may not be true if the change was remote — it depends on if the record was loaded locally before the change occurred.

In any case, it seems like we had different understandings of what #81 was about. I've investigated and the thing I thought that #81 would address is actually moot (I put a comment on that ticket explaining). You were thinking it was about automatically loading new records, which I think we agree now should be discussed further on #50. Okay to close this?

broerse:
Okay to close it.

PR #83, opened 2015-07-23
Ember-data never needs to reload ember-pouch records

rsutphin:
See #79 and #80 for discussions of why, and why this covers the individual record case only.


fsmanuel:
:+1:

PR #84, opened 2015-07-29
Change access to db property to use Ember.get

jkleinsc:
Right now in the ember adapter, the code is using this.db to reference the PouchDB used.  For most use cases this works fine, but it doesn't work if you are trying to set the db via injection doing something like this:
```javascript
var configDB = new PouchDB('config');
application.register('pouchdb:configdb', configDB, {instantiate: false});
application.inject('adapter', 'db', 'pouchdb:configdb');  
```
In the example above this.db will be undefined, but this.get('db') will return the defined PouchDB.

This pull request simply changes the this.db references to this.get('db') so that the PouchDB can be injected or can be used the way it is defined in the readme.

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -27,7 +27,7 @@ export default DS.RESTAdapter.extend({
   shouldBackgroundReloadRecord: function () { return false; },
 
   _startChangesToStoreListener: on('init', function () {
-    this.changes = this.db.changes({
+    this.changes = this.get('db').changes({
       since: 'now',
       live: true,
       returnDocs: false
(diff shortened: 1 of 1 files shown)


rsutphin:
@jkleinsc, thanks for this. It's a good idea. Can you rebase to get rid of that merge commit? If not I'll take care of it in the next few days.

jkleinsc:
@rsutphin I rebased and got rid of the merge commit.  Thanks!

rsutphin:
Thanks!

rsutphin:
Released as 2.0.3.

nolanlawson:
:+1:, thanks for the PR! :)

PR #86, opened 2015-08-25
Gives example of plugin import.

keppy:
This would have been helpful to me starting out, cheers


fsmanuel:
@keppy :+1:

nolanlawson:
Nice! Thanks a lot.

broerse:
There is no brocfile.js anymore. See:

https://github.com/ember-cli/ember-cli/blob/master/TRANSITION.md#brocfile-transition

PR #87, opened 2015-08-25
remove deprecation warnings for 1.13

### LINKED ISSUE #78: Deprecation warnings coming from the change watcher with ED 1.13
Reported by @OleRoel in #72:

> * Using store.getById() has been deprecated. Use store.peekRecord to get a record by a given type and ID without triggering a fetch
> * DS.Model#isDirty has been deprecated please use hasDirtyAttributes instead
> * You tried to look up 'store:main', but this has been deprecated in favor of 'service:store'

I think these are all coming from the change watcher.

fivetanley:
fixes #78

This change will be incompatible with the betas of Ember Data, just fyi. Users will have to upgrade Ember Data as well.

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -43,7 +43,7 @@ export default DS.RESTAdapter.extend({
     // skip changes for non-relational_pouch docs. E.g., design docs.
     if (!obj.type || !obj.id || obj.type === '') { return; }
 
-    var store = this.container.lookup('store:main');
+    var store = this.store;
 
     try {
       store.modelFor(obj.type);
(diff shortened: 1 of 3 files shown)


rsutphin:
@fivetanley, thanks for your attention to this. We are still trying to maintain backwards compatibility with the betas for the moment. Dropping backwards compatibility will require a major version bump.

I've tried eliminating the 1.13 deprecation warnings while preserving backwards compatibility. I was able to do it for all of them except for the `isDirty` check. Do know of a technique for that one?

rsutphin:
… thinking about it a little more, the event we were waiting for to drop backwards compatibility was, IIRC, ember-data releasing 1.0. Since that's happened now (and then some), maybe it's time for that major version bump. Anyone else have an opinion?

nolanlawson:
I'm not sure what percentage of Ember users have upgraded to 1.0, but I do see the value in getting rid of old cruft to deal with <1.0 versions. Agree with @rsutphin that it would be a breaking change, though.

broerse:
Perhaps we can use [ember-cli-version-checker](https://github.com/rwjblue/ember-cli-version-checker) to fix this?

rsutphin:
@broerse, we could do that, but we'd basically be maintaining two separate versions of the adapter then, just packaged together. Seems like if that's desirable we could just continue to maintain the 2.x line along with a new ED 1.13+-only 3.x line.

My guess is that there is not sufficient interest to maintain both lines, regardless of how they are packaged.

Does anyone have any objections to dropping compatibility with the ED betas?

broerse:
I am for dropping compatibility with the ED betas. But I think we must give an error like this:

https://github.com/mharris717/ember-cli-pagination/commit/6781cdef37ceb75ff328ea4f92d26b0d05abd35a

mattmarcum:
:+1: on dropping support for ED betas.

openhouse:
:+1: I needed this branch for Ember 2.2.0 Ember Data 2.2.1. Woo hoo! Now it's working!

rsutphin:
I've posted a PR (#95) which incorporates this one and officially drops compatibility with ember & ember-data earlier than 1.13.

PR #90, opened 2015-09-06
Created model blueprint

mattmarcum:
I use the  ember-cli model blueprint as basis so we don't reinvent the wheel.  The file template uses ember-pouch's model class.


rsutphin:
This looks good to me. Can you provide an example of how to use it in the README? (Sorry for the slow response.)

nolanlawson:
Ditto, would like an explanation of what this PR is doing. (Not too familiar with Ember blueprints.)

mattmarcum:
I updated the README with a section for blueprints and for this model blueprint.  I also added the section for my application adapter pr...so depending on the order of merging there will be a conflict.

Here's ember-cli's section on blueprints: http://ember-cli.com/extending/#generators-and-blueprints

fsmanuel:
I don't think it's a good idea to use `ember g model modelName` but I'm not an active ember-pouch user. What if I use ember-pouch for some models but have default ember-data models as well?

As I said I'm not an active ember-pouch user so you might know better.

rsutphin:
Thanks for the change, @mattmarcum. I didn't realize this was taking the plain "model" blueprint name — for some reason I thought it would be automatically namespaced by ember-cli. What do you think about `pouch-model` as the blueprint name instead? (Dasherized seems like the standard for ember-cli generators.)

If you agree, could you make that change and then rebase/squash? Thanks.

mattmarcum:
closing in favor of https://github.com/nolanlawson/ember-pouch/pull/101

PR #92, opened 2015-09-07
added attachment transform and unit test

mattmarcum:
This is just a commit of @stevebest's attachment transform from here: https://github.com/nolanlawson/ember-pouch/issues/37

### DIFF
--- a/addon/transforms/attachment.js
+++ b/addon/transforms/attachment.js
@@ -1,0 +1,28 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+
+export default DS.Transform.extend({
+  deserialize: function(serialized) {
+    if (Ember.isNone(serialized)) { return null; }
+
+    return Ember.keys(serialized).map(function (attachmentName) {
+      return {
+        name: attachmentName,
+        content_type: serialized[attachmentName]['content_type'],
+        data: serialized[attachmentName]['data']
+      };
+    });
+  },
+
+  serialize: function(deserialized) {
+    if (!Ember.isArray(deserialized)) { return null; }
+
+    return deserialized.reduce(function (acc, attachment) {
+      acc[attachment.name] = {
+        content_type: attachment.content_type,
+        data: attachment.data
+      };
+      return acc;
+    }, {});
+  }
+});
--- a/addon/transforms/attachment.js
+++ b/addon/transforms/attachment.js
@@ -5,7 +5,9 @@ export default DS.Transform.extend({
   deserialize: function(serialized) {
     if (Ember.isNone(serialized)) { return null; }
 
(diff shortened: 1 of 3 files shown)


addon/transforms/attachment.js
```diff
@@ -0,0 +1,28 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+
+export default DS.Transform.extend({
+  deserialize: function(serialized) {
+    if (Ember.isNone(serialized)) { return null; }
+
+    return Ember.keys(serialized).map(function (attachmentName) {
```
fsmanuel: `Ember.keys` is deprecated. You can do:

```
const keys = Object.keys || Ember.keys;

keys(serialized).map();
```

rsutphin:
Can you add an example of how to use this in the README? Otherwise LGTM.

nolanlawson:
LGTM, agree about adding documentation. Maybe someone else can take this up, since there are tests to demonstrate?

stevebest:
As an original author of this hack, I am obliged to say it misses a pretty critical piece in its current form: that is, handling of `stub: true` attachments. That was ok in case you were only attaching files once, or re-attaching all of them at once

Also, I found it more consistent to use `Ember.Object`s instead of just `Object`s.

That's how the hack looks like now:

```js
import Ember from 'ember';
import DS from 'ember-data';

const { isNone } = Ember;
const keys = Object.keys || Ember.keys;

export default DS.Transform.extend({
  deserialize: function(serialized) {
    if (isNone(serialized)) { return null; }

    return keys(serialized).map(function (attachmentName) {
      return Ember.Object.create({
        name: attachmentName,
        content_type: serialized[attachmentName]['content_type'],
        data: serialized[attachmentName]['data'],
        stub: serialized[attachmentName]['stub'],
      });
    });
  },

  serialize: function(deserialized) {
    if (!Ember.isArray(deserialized)) { return null; }

    return deserialized.reduce(function (acc, attachment) {
      const serialized = {
        content_type: attachment.get('content_type'),
      };
      if (attachment.get('stub')) {
        serialized.stub = true;
      } else {
        serialized.data = attachment.get('data');
      }
      acc[attachment.get('name')] = serialized;
      return acc;
    }, {});
  }
});
```

stevebest:
I'd be glad to provide it in a proper form, with tests. Shoud I start a separate PR?

nolanlawson:
Yes please! :)

mattmarcum:
@stevebest feel free to fork this branch if you want.  The unit tests in this branch should be a good starting point.

rsutphin:
@mattmarcum, thanks for kicking this off. @stevebest, please open a new PR when you've got the updated version ready.

PR #95, opened 2015-12-05
3.0: Support Ember/Ember-Data 1.13 and up only

rsutphin:
Builds on @fivetanley's PR #87 to drop support for Ember & Ember-Data before version 1.13. Includes changes to the ember-try & travis config to test on 1.13 and all 2.x versions so far.

Since this is a backwards-incompatible change, it bumps the package version to 3.0.

Given our limited test suite, I've normally been testing changes in my app that uses the adapter. However, that app is still on Ember 1.12, so I can't test these changes. I'd appreciate it if anyone who's interested could give it a try in their apps and let me know if there are problems.


broerse:
This looks good. Perhaps add https://github.com/rwjblue/ember-cli-version-checker for ember-data?

nolanlawson:
I'm all for dropping support for old versions, if it makes development easier. That said, we should probably have better tests. Also, is there a way to communicate via code which version of Ember Data we support? Something like npm `peerDependencies`?

mattmarcum:
:+1:  code looks good

broerse:
I get this error testing this branch.
```
Uncaught (in promise) CustomPouchError {status: 500, name: "unknown_error", message: "Database encountered an unknown error", error: true}
error: truemessage: "Database encountered an unknown error"
name: "unknown_error"
status: 500
__proto__: PouchError
constructor: PouchError(opts)
toString: ()
__proto__: Define
Error.bc
constructor: Error()
message: ""
name: "Error"
toString: toString()
__proto__: Object
```
I get this one time only and everything  seems to work. Not much time this week but will try to find this.

broerse:
https://github.com/broerse/ember-cli-blog/commit/438d2af39f7020032ebf35052a75a0d15e54ada0

Works!

fivetanley:
@nolanlawson beginning in Ember Data 2.4, Ember Data will be a full-fledged Ember add-on like Ember Pouch. You can use peerDependencies in that fashion. Until then, I think you would have to communicate this at runtime by asserting against `DS.VERSION`.

broerse:
@fivetanley @nolanlawson Is https://github.com/rwjblue/ember-cli-version-checker not the way to go?

rsutphin:
Okay, I added a version check. I used ember-cli-version-checker. It would have been cleaner to use a check against `DS.VERSION` since we wouldn't have to have knowledge of how ember-data is distributed at different versions. But I don't see a way to do that because I can't use `semver` (or other npm libraries) at runtime without a lot of complexity that seems unnecessary for a simple version check.

broerse:
@rsutphin The ember-data version check look good. Should we also add a version check for ember?

rsutphin:
I don't think we need a version check for ember. Ember-pouch mainly interacts with ember-data. It doesn't use any complex or subtle ember features — whatever version of ember works with the version of ember-data you have should be fine.

@broerse, thanks for checking in your app. Are there any further concerns before I merge this?

broerse:
@rsutphin It works OK for me but my app is still in 1.13.x. I see there are some issues with other modules and ember & ember-data 2.2.0. We could first double check 2.2.0.

I vote for merging now.

mattmarcum:
I just tried out this branch locally on ember js 2.2 and ember data 2.2.1 and everything worked fine.

broerse:
@mattmarcum Thanks for testing this!

rsutphin:
@mattmarcum, thanks for testing.

I tested with [HospitalRun](https://github.com/HospitalRun/hospitalrun-frontend). None of their tests that were passing before fail with this change. (I didn't run the full app, though.) I'm going to merge now.

Thanks for everyone's feedback.

rsutphin:
Published as 3.0.0.

PR #96, opened 2015-12-12
Use http instead of https

leo:
Their site doesn't yet have a working SSL certificate.


rsutphin:
I think it did at one point, but in any case the problem doesn't look like it's getting solved quickly.

PR #101, opened 2015-12-19
Added blueprint for pouch-model

mattmarcum:
This replaces https://github.com/nolanlawson/ember-pouch/pull/90

It creates a blueprint to easily create model files that import from `ember-pouch/model`

usage: `ember g pouch-model <model-name>`


rsutphin:
Thanks for sticking through the slow reviews and the change requests!

PR #103, opened 2015-12-19
Move blueprint files into explicit directories

backspace:
I just started a new `ember-pouch` application and used the `pouch-model` and `pouch-adapter` blueprints. They use the [`__path__` directory name](http://ember-cli.com/extending/#files):

> The `__path__` token is substituted with the blueprint name at install time. For example, when the user invokes ember generate controller foo then `__path__` becomes controller. When the `--pod` flag is used, for example `ember generate controller foo --pod` then `__path__` becomes `foo` (or `<podModulePrefix>/foo` if the podModulePrefix is defined). This token is primarily for pod support, and is only necessary if the blueprint can be used in pod structure. If the blueprint does not require pod support, simply use the blueprint name instead of the `__path__` token.

So when I ran `ember g pouch-adapter application`, it generated the adapter at `/app/pouch-adapters/application`. Likewise `ember g pouch-model destination` created `/app/pouch-models/destination.js`. Then Ember Data couldn’t find them!

I’m not sure whether this is the full solution because I know the `__path__` thing is involved in pods. But without pods, the way it’s generating files is broken!


mattmarcum:
Yeah, sorry originally when I made the blueprints they were named model/adapter and everything worked...but a name change was requested and I forgot to change the paths...good catch

rsutphin:
Agreed that we should take this as a solution for now — if someone's using pods and can assist getting them to work there also that would be great.

PR #106, opened 2016-01-15
Fix version check for Ember Data 2.3.x

jkleinsc:
Ember Data 2.3.x no longer needs to be included in bower.json (see http://emberjs.com/blog/2016/01/12/ember-data-2-3-released.html), but the version check in Ember Pouch does not support this.  This PR changes the version check to look for either NPM Ember Data > 2.2.99 or Bower Ember Data > 1.12.99.


jkleinsc:
I'm not sure why Travis CI is failing.  When I run the tests locally they pass.

backspace:
Maybe [this](https://github.com/ember-cli/ember-cli/issues/5316)?

jkleinsc:
@backspace my commit includes pinning jQuery to 1.11.3, so I don't think that is the problem.  I think the issue may be that the version checker code doesn't quite work as expected with ember-try.  I am going to check in a fix that should address the issue.

backspace:
Oh, sorry, I missed that!

jkleinsc:
I found and fixed the issue and have squashed my commits, so this PR is ready to go.

nolanlawson:
Awesome, LGTM. Thank you!

nolanlawson:
@jkleinsc you're now a collaborator. :) The rules are pretty simple: open a PR and wait for someone to review and give a +1, unless it's something trivial like a README change - then just go ahead and merge it. Or timebomb it if nobody reviews after a few days. Welcome and thanks again!

PR #108, opened 2016-01-22
Make adapter call a hook when encountering a change for a record that is not yet loaded.

lukemelia:
Pouch adapter now calls a hook method when encountering a change for a record that is not yet loaded.

- The hook is no-op by default, to match present behavior and performance profile
- Added a test demonstrating use of the hook to load the new data into the ember-data store

@lukemelia / @chrislopresto

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -76,6 +76,7 @@ export default DS.RESTAdapter.extend({
     var recordInStore = store.peekRecord(obj.type, obj.id);
     if (!recordInStore) {
       // The record hasn't been loaded into the store; no need to reload its data.
+      this.unloadedDocumentChanged(obj);
       return;
     }
     if (!recordInStore.get('isLoaded') || recordInStore.get('hasDirtyAttributes')) {
@@ -90,6 +91,19 @@ export default DS.RESTAdapter.extend({
     } else {
       recordInStore.reload();
     }
+  },
+
+  unloadedDocumentChanged: function(/* obj */) {
+    /*
+     * For performance purposes, we don't load records into the store that haven't previously been loaded.
+     * If you want to change this, subclass this method, and push the data into the store. e.g.
+     *
+     *  let store = this.get('store');
+     *  let recordTypeName = this.getRecordTypeName(store.modelFor(obj.type));
+     *  this.get('db').rel.find(recordTypeName, obj.id).then(function(doc){
+     *    store.pushPayload(recordTypeName, doc);
+     *  });
+     */
   },
 
   willDestroy: function() {
--- a/README.md
+++ b/README.md
@@ -137,7 +137,17 @@ Out of the box, ember-pouch includes a PouchDB [change listener](http://pouchdb.
 
 However, ember-pouch does not automatically load new records that arrive during a sync. The records are saved in the local database, but **ember-data is not told to load them into memory**. Automatically loading every new record works well with a small number of records and a limited number of models. As an app grows, automatically loading every record will negatively impact app responsiveness during syncs (especially the first sync). To avoid puzzling slowdowns, ember-pouch only automatically reloads records you have already used ember-data to load.
 
-If you have a model or two that you know will always have a small number of records, you can write your own change listener to tell ember-data to automatically load them into memory as they arrive.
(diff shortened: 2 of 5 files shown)


addon/adapters/pouch.js
```diff
@@ -92,6 +93,19 @@ export default DS.RESTAdapter.extend({
     }
   },
 
+  unloadedDocumentChanged: function(/* obj */) {
+    /*
+     * For performance purposes, we don't load records into the store that haven't previously been loaded.
+     * If you want to change this, subclass this method, and push the data into the store. e.g.
+     *
+     *  let store = this.get('store');
+     *  let recordTypeName = this.getRecordTypeName(this.store.modelFor(obj.type));
```
fsmanuel: `store.modelFor(obj.type)` without `this`?

nolanlawson:
Hmmm, reading this and trying to understand it (forgive me; I am really weak on EmberData specifics, although I grok Pouch and relational-pouch :smiley:).

I guess what I'm wondering is why we have [this early return](https://github.com/nolanlawson/ember-pouch/blob/25c3866dad9633e6805a72fd8a29bdef1978890d/addon/adapters/pouch.js#L77-L79) in the first place. Doesn't this mean that changes from a remote CouchDB will never be loaded into the Ember store? Why don't we just make your commented code the default behavior?

(Side note: it seems like we have a very robust syncing system between PouchDB and CouchDB, but a rather weak one between PouchDB and the in-memory store. In particular, the fact that this syncing logic isn't taking `_conflicts` or `_revisions` into consideration worries me a lot. I'm not sure if anything's wrong, but it just seems like a code smell to me.)

fsmanuel:
@nolanlawson here is the discussion: https://github.com/nolanlawson/ember-pouch/issues/50#issuecomment-86102868

lukemelia:
@nolanlawson @fsmanuel yes, my understanding based on the README and the referenced thread was that pushing every new record into the store on arrival was an undesirable for performance reasons. We have a low-volume use case and wanted to make it easier to implement that approach.

rsutphin:
@nolanlawson, I know you're busy so here's a summary: the ember-data store is the in-memory cache of records the application has touched during the current session. For records which are already cached, our default change watcher updates the cached copy when they change and removes them from the cache when they are removed from PouchDB. It does not automatically cache new records that arrive because that causes terrible performance during the initial sync with medium-and-up data sizes.

(Re: revisions. The change watcher does not update cached records unless ember-data's state machine believes that the cached record is up to date with PouchDB; see [this line](https://github.com/nolanlawson/ember-pouch/blob/0b5e8b8e3b39be2a92599b9b6b813c715e8220c8/addon/adapters/pouch.js#L81). If an in-memory copy has been modified but not saved and simultaneously a sync brings in a new rev of that record, the application's attempt to save it will cause a conflict. The application can then resolve the conflict as it sees fit.

(Similarly for `_conflicts`: if there are conflicts during replication, it's the responsibility of the app to take care of them in PouchDB directly. When the app resolves them, the change watcher will update any loaded records. If the app ignores `_conflicts`, I guess that the change watcher would receive changes that reflect the default resolutions, but I haven't tried that myself.)

rsutphin:
@lukemelia / @chrislopresto: thanks for this. I have an idea about how to make the adapter act  more naturally with regards to this area, but I haven't had a chance to implement/propose it yet. Your change would not interfere with what I'm thinking of, and it looks like a good interim step.

lukemelia:
Thanks for the summary @rsutphin. With respect to the discussion of conflict handling. This seems like it could also be initially addressed by calling a hook with the relevant details that has no-op behavior. i.e. but the power in the hands of the end-user while providing conservative defaults.

nolanlawson:
@rsutphin Ah okay, that makes a lot more sense then. I look forward to the more complete in-memory mirroring solution (I take it something [like this](http://pouchdb.com/2015/02/28/efficiently-managing-ui-state-in-pouchdb.html) would reproduce exactly the performance problem you describe, for the initial load of a large database?), but for now, if you're +1 on this PR, then I am as well. I'll let you merge it when you're ready. :)

mattmarcum:
:+1: is there anything that needs to be done to get this merged?

nolanlawson:
I just wanted to give other contributors time to look at it. I'm +1; I'll merge later today when I get a chance.

Issue #16, opened 2014-11-20
Bidirectional hasMany relationships are not loadable/saveable from the many side


rsutphin:
The default behavior of `DS.JSONSerializer` is to not serialize the many side of bidirectional hasMany relationships. See emberjs/data#2494 for links to discussion about this.

The upshot for `ember-pouch` is that the contents of a bidirectional hasMany relationship isn't saved or loaded when working from the many side. E.g., say you have these models:

```javascript
// app/models/person.js
export default DS.Model.extend({
  sandwiches: DS.hasMany('sandwich', { inverse: 'owner' })
});

// app/models/sandwich.js
export default DS.Model.extend({
  owner: DS.belongsTo('person')
});
```

In this case you won't be able to load or save a person's sandwiches from the person. (You will be able to if you remove `owner` from sandwich, though.)

The workaround for this that other similar adapters use ([ember-localstorage-adapter][LS], [ember-indexeddb-adapter][IDB]) is to provide a serializer which forces these associations to be serialized on the hasMany side (as well as the belongsTo side). I'm doing this now in my own application to work around this issue.

[LS]: https://github.com/kurko/ember-localstorage-adapter/blob/c32477626ce0e7c8447e1ee79eb78e64a8357d70/localstorage_adapter.js#L6
[IDB]: https://github.com/kurko/ember-indexeddb-adapter/blob/0b1c73a1134606ccc9dfc1aadb86af5b4dd9fe4d/packages/indexeddb-adapter/lib/indexeddb_serializer.js#L2

Unfortunately this requires copying a bunch of code from JSONSerializer (hence emberjs/data#2494). It also duplicates the information about the relationship, requiring both sides to be saved manually whenever the relationship changes.

I'm not sure if there's a practical better solution. It appears that the ember-data solution is to have the server use different serializations for reads vs. writes, but that doesn't work for a local database. One idea I had was that the adapter could look up the IDs for the hasMany relationship dynamically — since all the data is local this might work, but it also might be complex to get right.

fsmanuel:
I'm not sure when ember-data added `_shouldSerializeHasMany` but in 1.13.x it is available
We can add that line to the serializer and add a section to the readme that ppl should import the serializer.

```js
  _shouldSerializeHasMany: function() { return true; }
```

broerse:
Or use an initializer to import it? Example: https://github.com/firebase/emberfire/blob/master/addon/initializers/emberfire.js

[...]

PR #109, opened 2016-01-23
Add failing test for loading hasMany associations

backspace:
This incorporates the proposed fix for #16, but it doesn’t
actually fix the problem. Perhaps I’m misunderstanding!


nolanlawson:
Sounds like you've demonstrated that the proposed fix doesn't actually fix anything? In which case, this is a very useful PR. :)

backspace:
Yeah! I’m curious to hear from other people in that thread though because multiple people seemed to be having success with it.

PR #111, opened 2016-01-31
Add override so serialiser saves hasMany

backspace:
As suggested [here](https://github.com/nolanlawson/ember-pouch/issues/16#issuecomment-122651303) by @fsmanuel.

What I missed in my [previous pull request](https://github.com/nolanlawson/ember-pouch/pull/109) was that the dummy application wasn’t using the PouchDB serialiser!

### DIFF
--- a/tests/dummy/app/serializers/application.js
+++ b/tests/dummy/app/serializers/application.js
@@ -1,0 +1,3 @@
+import { Serializer } from 'ember-pouch';
+
+export default Serializer;
--- a/addon/serializers/pouch.js
+++ b/addon/serializers/pouch.js
@@ -1,3 +1,19 @@
 import DS from 'ember-data';
 
-export default DS.RESTSerializer.extend({});
+export default DS.RESTSerializer.extend({
+  _shouldSerializeHasMany: function() {
+    return true;
+  },
+
+  // This fixes a failure in Ember Data 1.13 where an empty hasMany
+  // was saving as undefined rather than [].
+  serializeHasMany(snapshot, json, relationship) {
+    this._super.apply(this, arguments);
+
+    const key = relationship.key;
+
+    if (!json[key]) {
+      json[key] = [];
+    }
+  }
+});
--- a/tests/dummy/app/models/food-item.js
+++ b/tests/dummy/app/models/food-item.js
@@ -5,5 +5,6 @@ import DS from 'ember-data';
 export default DS.Model.extend({
   rev: DS.attr('string'),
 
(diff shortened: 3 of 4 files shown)


tests/dummy/app/serializers/application.js
```diff
@@ -0,0 +1,3 @@
+import { Serializer } from 'ember-pouch/index';
```
fsmanuel: you don't need index:
`import { Serializer } from 'ember-pouch';`

fsmanuel:
The failing test is related to the change:

`Integration | Adapter | Basic CRUD Ops: update an existing record`
> Error: Assertion Failed: A (subclass of DS.Model) record was pushed into the store with the value of ingredients being 'undefined', but ingredients is a hasMany relationship so the value must be an array. You should probably check your data payload or serializer.

I don't know if it's a bug in ember-data 1.13.

tests/dummy/app/serializers/application.js
backspace: Ah, thanks. I emulated [this line](https://github.com/nolanlawson/ember-pouch/blob/master/tests/dummy/app/adapters/application.js#L1). Changed now.

backspace:
@fsmanuel, I was able to get the tests passing on Ember 1.13 by adding this override of `serializeHasMany`:

```
  serializeHasMany(snapshot, json, relationship) {
    this._super.apply(this, arguments);

    const key = relationship.key;

    if (!json[key]) {
      json[key] = [];
    }
  }
```

If the Ember Data implementation of `serializeHasMany` would result in an `undefined` attribute, this replaces it with an empty array. It seems kind of weird to me, but it works… do you think I should add this to the PR?

fsmanuel:
@backspace LGTM!

nolanlawson:
Seems to be failing only in Ember Data 1.13:

> A (subclass of DS.Model) record was pushed into the store with the value of ingredients being 'undefined', but ingredients is a hasMany relationship so the value must be an array. You should probably check your data payload or serializer.

Do we want to drop support for 1.13 or is this fixable?

backspace:
@nolanlawson: I just pushed that override for `serializeHasMany` that [fixes it on 1.13](https://travis-ci.org/nolanlawson/ember-pouch/jobs/106208731). It’s a bit awkward but seems harmless.

nolanlawson:
LGTM! Thanks for the fix; passing in Travis now.

nolanlawson:
3cb223173ca53c3a864ae23db86c79b43c020fa9

PR #122, opened 2016-03-19
Serializer blueprint

mattmarcum:
The [serializer/relations bug](https://github.com/nolanlawson/ember-pouch/issues/116) bit me in the ass.  Couldn't find anything in the Readme that mentioned the need for a serializer so I added a blueprint and updated the readme.


nolanlawson:
Thanks for the PR! Could somebody else with more Ember knowledge take a look at this? Also just to check - this wasn't a bug that snuck in in v3.2.0, was it?

mattmarcum:
I don't think it was a bug so much as ember-pouch doesn't act like you'd expect w/r/t relations.  I think this https://github.com/nolanlawson/ember-pouch/pull/111 fixed it, but then you'd need to create an application seralizer in your app for it to work.  I just didn't see it documented anywhere that you really needed a serializer.

backspace:
I agree that this should be mentioned in the documentation, so I support merging this.

I also wonder whether we should make explicit that people need to save both ends of a relationship because that’s not how people will generally expect Ember Data to work.

nolanlawson:
Is this related to https://github.com/nolanlawson/ember-pouch/issues/5?

dweremeichik:
As far as making it explicit that both ends need to be saved, that is in the documentation now under relationships - > saving.

fsmanuel:
@backspace why do you think that ember-data users expect that they only need to save one side? Have I missed something?

backspace:
Oops, missed that, @dweremeichik, thanks!

@fsmanuel, with Ember Data don’t you normally only have to save the child end of a relationship because in a relational database-style API, it’s the one with the foreign key?

dweremeichik:
@fsmanuel @backspace according to the Emberjs documentation you only need to save one side. It shouldn't matter which, see [this](https://guides.emberjs.com/v2.4.0/models/relationships/#toc_creating-records). Anyone who is new to both Ember and Ember Pouch will assume that what the documentation says is what would be done here (as it was never specified otherwise). Per the recent ember pouch documentation update, this is made more clear for new users. I would agree that it makes sense to include the serializer by default, otherwise it won't work as users expect.

fsmanuel:
@backspace @dweremeichik  yeah that's true. But it only works for relational database-style API. As far as I know ember data doesn't save both sides (aka send 2 requests to the server).
For ember-local-storage I added a `autoSave` [option](https://github.com/funkensturm/ember-local-storage/blob/master/addon/adapters/adapter.js#L50) to the relationship to autoSave the other side.

If someone is interested I can assist to bring that feature into ember-pouch.

backspace:
That sounds great to me, @fsmanuel!

mattmarcum:
Does this PR need any changes in order to get merged?

backspace:
I ran your generator locally and it worked as expected. I think @nolanlawson prefers squashed commits so maybe do that and then it can be merged?

mattmarcum:
closing this in favor of https://github.com/nolanlawson/ember-pouch/pull/124

PR #124, opened 2016-03-30
Added serializer blueprint and updated docs

mattmarcum:
This pr is a squashed version of https://github.com/nolanlawson/ember-pouch/pull/122

It adds a serializer blueprint and updates the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ This provides
 - `import PouchDB from 'pouchdb'`
 - `import {Model, Adapter, Serializer} from 'ember-pouch'`
 
+You must install an adapter and serializer in order for `Ember-Pouch` to work correctly.  The [`blueprints`](#emberpouch-blueprints) are the easiest way to get started.
+
 `Ember-Pouch` requires you to add a `rev: DS.attr('string')` field to all your models. This is for PouchDB/CouchDB to handle revisions:
 
 ```js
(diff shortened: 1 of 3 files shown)


nolanlawson:
Sorry I haven't had a chance to look at this PR, could someone with more experience with Ember blueprints look at this please?

fivetanley:
@mattmarcum I think a `serializers/application.js` in the blueprints would be a great start. Users should get prompted if there is a conflict. But having it in the README could make it a first step that could lead to confusing bugs if missed.

nolanlawson:
@fivetanley So you mean this PR would make more sense without the README modification?

fivetanley:
We ran into this recently with the ActiveModelAdapter (this reminds me I need to fix it there too). Basically, if you don't define a serializer for a model, it will look for `ApplicationSerializer`, and then the default serializer for the adapter (in 2.0 this is `JSONAPISerializer`, 'RESTSerializer` in 1.x). The reason why the README modification was made (I'm guessing) is to account for this behavior. However, if we make having the PouchDB Adapter and Serializer be the default, this would do the right thing for users instead of asking them to correct this. This may be something we rework for Ember Data 3.0, but we can work around it in addons for now.

fivetanley:
As somebody who usually skims for the installation step and then starts hacking, I think making the application.js serializer/adapter is the move that sets users up for success.

backspace:
I was reluctant to make it the default application serialiser/adapter but your [point](https://github.com/nolanlawson/ember-pouch/pull/124#issuecomment-210901591) that it would prompt if there’s a conflict is true… so yeah, I also support adding it to the blueprint.

aexmachina:
@mattmarcum, thanks for this PR. I think the generator is the right approach - we actually hit this issue ourselves. It looks like we just need to update the README to direct people to use the generator and this can be merged.

broerse:
@aexmachina Can you make the readme change or close the PR?

PR #140, opened 2016-09-14
Added option to disable live sync and support for periodic polling for changes

aexmachina:


aexmachina:
@broerse I'd like to get this PR merged, can you please review?

broerse:
@aexmachina I understand the need to specify the [Change Notifacation](http://guide.couchdb.org/draft/notifications.html) type. But I don't think it should be a enable/disable flag. CouchDB has now 3 options, It could be more in the future. I also think this should be handled by pouchdb and from ember-pouch we just pass the option to pouchdb. So I am not sure we should merge this in this form.

aexmachina:
I probably should have been clearer about the intent of this PR. The goal is to allow an adapter to use a PouchDB instance that has an HTTP transport without using long polling. Currently, if you provide a PouchDB instance with an HTTP transport (i.e. directly to the remote - without a local) then the adapter's call to `db.changes()` will use long polling, which is a problem for us because we end up running out of sockets when multiple tabs are opened.

This PR provides two features to solve this:

1. `liveSync: false` disables the long polling
2. `syncInterval: 10000` polls for changes every 10 seconds

broerse:
@nolanlawson I think it is better to add something like a:
```
feed=normal
feed=longpoll
feed=continuous
feed=eventsource
```
option to pouchdb and we set this option in ember-pouch. Perhaps only support `normal` , `longpoll` , `eventsource` for now. What do  you think?

aexmachina:
@broerse I agree that is a better solution, if @nolanlawson is open to adding support to PouchDB then I can look at doing a PR. Could take me a while to get to it though!

I'm thinking that adding a `feed` option to the `#changes` method would be the right way to expose this. Probably deprecating the `live: true` option and directing people to use `feed: 'longpoll'` instead.

broerse:
`feed:` and `live:` are not the same. Live replication (or "continuous" replication) is a separate mode where changes are propagated between the two databases as the changes occur. In other words, normal replication happens once, whereas live replication happens in real time. So this does not need to be deprecated if we introduce `feed:`

aexmachina:
@broerse `live: true` would be equivalent to `feed: 'longpoll'`. Remember that in order to move this functionality to PouchDB it does need to be continuous i.e. repeatedly polling for changes or maintaining a long polling, continuous or SSE connection.

broerse:
@aexmachina with `live: true`, `feed: 'longpoll'` is "continuous" replication. With `live: false` , `feed: 'longpoll'` replication happens once. Also with `live: true`, `feed: 'normal'` is "continuous" replication. With `live: false` , `feed: 'normal'` replication happens once. So we don't agree on that this is equivalent.

aexmachina:
Thanks for clarifying that @broerse, now I understand better. The `feed` option doesn't seem to be documented anywhere. It seems to me that the options interface could be better because `live: false` doesn't really make sense with any `feed` type other than `normal`. Hence my suggestion that `live` be replaced with `feed`. Anyway, I'm happy to take whatever guidance you or @nolanlawson want to offer.

On that note, @nolanlawson doesn't seem to be replying. Is there someone else I should be speaking to about this?

broerse:
@aexmachina Perhaps you should open an issue on pouchdb asking for `feed: 'longpoll'` and `feed: 'normal'` support.

Issue #141, opened 2016-09-19
Move Relationships Copy and Paste Code to Adapter


zaubererty:
Are there any plans to move the relationship code to the adapter?

Cause I'm used to the way it's handled by other ember-data adapters like ember-local-storage or the default json-api adapter.

All relations are saved, deleted and side loaded automatically.

fsmanuel:
@zaubererty you can "steal" my code from ember-local-storage and open a PR if you want.

zaubererty:
@fsmanuel I'll look into it

aexmachina:
The relationship handling in ember-pouch is provided by relational-pouch, and my view is that this is the correct layering. The relationship handling that's provided by relational-pouch is not specific to ember-data, and it's a benefit to have this available to people who want to use PouchDB in this way outside of ember-data.

dehuszar:
That doesn't mean that ember-pouch is a bad place to extend/alter that relationship handling for Ember-specific purposes...

fsmanuel:
@dehuszar it's more like relationship sugar than actually relationship handling. It would eliminate the [relationships section](https://github.com/nolanlawson/ember-pouch#relationships) in the readme.

fsmanuel:
@zaubererty the relevant part is [here](https://github.com/funkensturm/ember-local-storage/blob/master/addon/adapters/base.js#L39-L87).

aexmachina:
@dehuszar I understand what you're suggesting now, and yes I think that's reasonable. I think it's important that the behaviour is opt-in, as it is in @fsmanuel's example code.

dweremeichik:
I submitted a pull request (#145) to do the deletion portion of this.
(note that I have not tested this in production, but the two tests I wrote appear to work.)

I am however, having trouble with the creation portion. When saving the parent record I am getting the following error `Cannot read property '_relationships' of undefined`. From watching this value through the tests, it looks like this should actually be set to an array with the model of the child. @fsmanuel did you run into anything like that when you implemented it on ember-local-storage?

fsmanuel:
@dweremeichik When do you get the error?

zaubererty:
I have seen the same issue. 
For me the reason is that the saving part is not completed yet when you try to add the ID of the related after calling 'rel.save'.

```javascript
 createRecord: function(store, type, record) {
    this._init(store, type);
    var data = this._recordToData(store, type, record);
    return this.get('db').rel.save(this.getRecordTypeName(type), data).then(
[...]

PR #145, opened 2016-10-12
Added cascading deletion functionality to hasMany relationships.

dweremeichik:
This is a partial fix to #141. 
This adds functionality for cascading deletion of children in a `hasMany` relationship.

Tests added, README updated.
Removed utils.js as we now return an empty promise upon successful deletion, instead of null. :see_no_evil:


addon/adapters/pouch.js
```diff
@@ -363,10 +359,33 @@ export default DS.RESTAdapter.extend({
...
-    return this.get('db').rel.del(this.getRecordTypeName(type), data)
-      .then(extractDeleteRecord);
+    let data = this._recordToData(store, type, snapshot);
+
+    //Delete the children
+    let deletedChildren = [];
+    snapshot.eachRelationship(function(name, descriptor) {
+      if(descriptor.kind === 'hasMany' && descriptor.options['cascadeDestroy']) {
+        snapshot.record.get(name).then(function(childRecords) {
+          childRecords.forEach(function(childRecord) {
```
srsgores: I feel like you could use `map` here, to map the `destroyRecord()` instead of using `forEach`.

```
childRecords.map((item) => item.destroyRecord());
```

[The former will be faster than both `forEach` **and** `for`](https://ryanpcmcquen.org/javascript/2015/10/25/map-vs-foreach-vs-for.html).

addon/adapters/pouch.js
```diff
@@ -363,10 +359,31 @@ export default DS.RESTAdapter.extend({
...
-    var data = this._recordToData(store, type, record);
-    return this.get('db').rel.del(this.getRecordTypeName(type), data)
-      .then(extractDeleteRecord);
+    let data = this._recordToData(store, type, snapshot);
+
+    //Delete the children
+    let deletedChildren = [];
+    snapshot.eachRelationship(function(name, descriptor) {
+      if(descriptor.kind === 'hasMany' && descriptor.options['cascadeDestroy']) {
+        snapshot.record.get(name).then(function(childRecords) {
```
fsmanuel: I think it should be:
```js
deletedChildren = snapshot.record.get(name).then(function(childRecords) {
  return childRecords.map(childRecord => childRecord.destroyRecord());
});
```

dweremeichik: You are right, I did mess something up. What you have is pretty close, it should be fixed in the latest commit.

addon/adapters/pouch.js
```diff
@@ -367,9 +367,10 @@ export default DS.RESTAdapter.extend({
     let deletedChildren = [];
     snapshot.eachRelationship(function(name, descriptor) {
       if(descriptor.kind === 'hasMany' && descriptor.options['cascadeDestroy']) {
-        snapshot.record.get(name).then(function(childRecords) {
-          deletedChildren = childRecords.map(childRecord => childRecord.destroyRecord());
+        let mappedPromise = snapshot.record.get(name).then(function(childRecords) {
```
srsgores: Can be optimized with arrow function:

```
let mappedPromise = snapshot.record.get(name).then(childRecords => Ember.RSVP.all(childRecords.map(childRecord => childRecord.destroyRecord())));
```

addon/adapters/pouch.js
```diff
@@ -365,9 +365,9 @@ export default DS.RESTAdapter.extend({
 
     //Delete the children
     let deletedChildren = [];
-    snapshot.eachRelationship(function(name, descriptor) {
+    snapshot.eachRelationship((name, descriptor) => {
       if(descriptor.kind === 'hasMany' && descriptor.options['cascadeDestroy']) {
-        let mappedPromise = snapshot.record.get(name).then(function(childRecords) {
+        let mappedPromise = snapshot.record.get(name).then(childRecords => {
```
srsgores: Personal preference, but I might actually go with the parentheses for params here, just to be consistent with [line 368](https://github.com/nolanlawson/ember-pouch/pull/145/files/cd46f568efccb5c9d0214281162535edecc13189..eee182892d8b2c2977e5c392381bb86b0d3e2a16#diff-8bef95280660598dc2f4437f447292e2R368):

```
let mappedPromise = snapshot.record.get(name).then((childRecords) => {
```

Preference, not necessary.

dweremeichik: Ha... I do that in multiple places... You do raise a good point here about this repo and consistency. I just created a new issue #149 for a contributing guide. Perhaps preferences like this could be defined in there.

broerse:
@fsmanuel Can I merge this?

dweremeichik:
@broerse @fsmanuel  I would hold off, @zaubererty was looking to do a combined pull request with both the auto save and cascading deletion functionality.

jlami:
Shoudn't deleteRecord only mark the ember record as deleted? And only after a save should it be really deleted on the db side?

http://emberjs.com/api/data/classes/DS.Model.html#method_destroyRecord

dweremeichik:
@jlami no, it should not, as we are more or less trying to stay compatible with[ ember-local-storage](https://github.com/funkensturm/ember-local-storage/blob/master/addon/adapters/base.js#L60-L87) on this one. Think of this functionality as being similar to a [cascade delete in SQL](http://www.mysqltutorial.org/mysql-on-delete-cascade/). This functionality is opt in, and is likely a good choice in many cases. If there are cases where you don't want to have it indiscriminately delete all of the children, you can still handle it yourself. I suppose if there was a large enough need for it, we could look into making another flag that would have it behave as you describe.

@broerse @srsgores @fsmanuel Thoughts? Have you guys run into many situations where what @jlami is describing has been needed?

broerse:
@dweremeichik I think the ember way is to only mark the record as deleted. So just as @jlami I am not sure this is correct. 

deleteRecord
---------

> Marks the record as deleted but does not save it. You must call save afterwards if you want to persist it. You might use this method if you want to allow the user to still rollbackAttributes() after a delete it was made.

dweremeichik:
@broerse sure, but we are in essence talking about two different things here. The `deleteRecord` both @jlami and yourself are talking about is for **DS.Model**. [`deleteRecord` in  **DS.Adapter**](http://emberjs.com/api/data/classes/DS.Adapter.html#method_deleteRecord) actually handles the deletion of the record on the server, which is what I am doing here. That is why it calls into relational-pouch with the `rel.del`, how this part of the adapter has always behaved. 

The new portion of the code simply allows you to automatically delete all child records when you delete a parent record. Why would you need to approve the deletion of the child records (which are useless without the parent) when you have already told the adapter to delete the parent by either calling `destroyRecord` or `deleteRecord` and then `save` on the parent record? If that was the case, there is no real benefit in this PR as you could simply loop over the records children yourself and mark them with DS.Model's `deleteRecord`. As our README.md already advises the user to do, this is simply a shortcut, or sugar for the developer. Again, our code here is really not that different from [this snippet](https://github.com/funkensturm/ember-local-storage/blob/master/addon/adapters/base.js#L60-L87) in ember-local-storage.

@jlami if I am misunderstanding your initial comment, please correct me.

fsmanuel:
@dweremeichik is totally right that the `deleteRecord` on a model is something else. As he said its more sugar. In Ruby on Rails you can define this behavior as well and that's what I implemented in ember-local-storage. I think it's a good idea. I don't use ember-pouch and someone else should review the implementation but it looks good to me.

broerse:
@fsmanuel Get your point. Thanks!

jlami:
@dweremeichik You're right. I guess the naming is a bit confusing. If the destroyRecord on the model is the real final removal, why not call the method on the adapter the same.

But I'm also a bit reluctant to have this kind of "magic deletion" occur in ember-pouch. If it was really helpful, shouldn't it be in the ember-data side? I don't really see why this should be pouch specific. I don't say it is really a bad idea, but maybe not the place or the right fix for the problem.

I'm trying to weigh in here because I'm trying to make ember-pouch and relational-pouch not save the hasMany part. This might also fix the problem of #141 as far as I can tell

This has the benefit of not only work more like normal ember data, in that you only need to save the child side of the one-to-many relation as is "normal" in relational databases. But it also is more in line with couch as I see it. That is because the tracking of the hasMany ids makes this all very heavy. In my case I got extra conflicts even. With only tracking the belongsTo side of the relationship you get more in line of the couch way of adding/removing in my mind. Simple, small, clean transactions, that don't really need any modifications on a global 'sum'.

I hope to put in a PR for these lighter relationships this weekend so you can check it out too.

jlami:
I think I have changed my mind on whether this is useful or not. This feature is the db equivalent of the cascade options. I do not know if ember-pouch is the best place for this, as this feature could probably be shared across a lot of serverless db-stores, but for now this is probably the right place.

One extra concern this PR could probably address is the other options for relationship reference checks. The 'cascade to null' and 'error/prevent deletion' options that you have in other databases.

The cascade to null is probably the easiest, ember does this by default if it can't find a record that is referenced by a belongsTo. Ember-pouch does have a 'bug' there now, but this is easily fixed. `adapter.findRecord` should not throw an Error when relational-pouch does not find anything. But it should fail the promise with `return Ember.RSVP.reject();`

Preventing the delete from going through when there are still entities referencing this record might also be a simple change to your current code. Instead of deleting the child records. You only search for any and then throw an error or reject the promise, which ever is appropriate here.

What do you think?

broerse:
> Preventing the delete from going through when there are still entities referencing this record

Would perhaps need a flag to choose between `cascading delete` and `delete when not referenced`. I would like to have the `delete when not referenced` option. Should this be a global flag?

dweremeichik:
Closing as stale.

PR #147, opened 2016-10-12
Updated ember-cli fixes and some minor changes

Fed03:
I've updated the addon to the latest ember-cli version available and other minor changes.

This PR is preparatory to some structure changes (not API) I've in mind that could open to a plugin ecosystem. For example I'm developing an `ember-simple-auth` plugin that uses `pouchdb-authentication` and hooks into `ember-pouch`.

I thought to open this PR first to know your opinions and to prepare the ground without an excessive shock


broerse:
Still Beta but take a look at https://github.com/martinic/ember-simple-auth-pouch used like this https://github.com/broerse/ember-cli-blog/blob/master/app/authenticators/pouch.js

broerse:
It looks good. Perhaps `git squash` this in one commit?

Fed03:
@broerse squash done!
About the existing beta addon I took a look at it and seems pretty good. As a matter of fact, I'm working over it and will do a PR there as well.
Lemme know if you are interested in the structure changes and/or where to propose them

broerse:
Yes please improve the beta addon.  I am interested in structure changes. I like to see the move to NPM only. See also https://github.com/nolanlawson/ember-pouch/issues/120

broerse:
@Fed03 It seems you must also change https://github.com/Fed03/ember-pouch/blob/new-version/blueprints/ember-pouch/index.js versions.

Fed03:
@broerse I kept the current bower deps cause I didn't want to change anything that "matters" in this PR; but, if you want, I'll update them.

I don't really thought to switch to npm only. I read that if we want to use npm module in ember we must use something like `ember-browserify`, but I'm not so sure there's the only way. For ex, `ember-data`moved from bower to npm altough they basically rewritten it in an ember-addon way. 
Maybe I could use something like rollup during the compilation time.
I don't know how to do it right now, think you could give me any suggestion? thnx

broerse:
@Fed03 I will merge it like this.

PR #150, opened 2016-10-13
Added CONTRIBUTING.md

dweremeichik:
Added CONTRIBUTING.md. Moved environment setup from README.md to the contributing doc. Link to contributing added to README.md. Note that this should also show up in link form when someone creates a new issue or pull request.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributing | Ember Pouch
+___
+Hey, thanks for making a difference in the Ember Pouch community!
+
+There are many ways you can contribute to this wonderful community.
+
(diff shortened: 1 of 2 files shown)


broerse:
I am against moving how to install from the readme. I like it that everybody is contributing in their own way. Lets not start telling people what to do. So I am for closing this PR.

dweremeichik:
I suppose i didn't look at it from that point of view. I was looking at it more from the angle of helping newcomers in the community. After all the ember community is highly opinion based. 

I wouldn't say that the contributing doc tells anyone explicitly what to do other than possible good practice. In fact it encourages coming up with new ways to help out. 

Perhaps instead of killing it we modify it?

Issue #120, opened 2016-03-09
Move 'pouchdb' to its own ember-cli addon


fivetanley:
Sometimes, I need to do something that doesn't use Ember Data or implement my own Ember Data adapter due to reasons™. 

Here's some stuff I've been working on that the separation would help:

1) Fastboot, it's easier to manage the pouch source when it is its own plugin rather than continuing to add shim files. I'm working on an addon that does this work for you.
2) A JSONAPI-ish adpater for PouchDB (writes jsonapidocs instead of ember-relational).
3) It would be nice to be able to load pouchdb-relational separately via an `import` statement instead of having it always included if possible.

This wouldn't be a breaking change, as `ember-pouch` (this addon) could rely on the new addon . Alternatively, the ember data stuff in this addon could be moved to a new addon instead.

As always, thanks for everything. Pouch is a great ecosystem.

broerse:
Ember is moving away from bower. I don't think pouchdb needs to be an `ember-cli` addon but it would be logical for `ember-pouch` to move to the NPM versions of `pouchdb' and `relational-pouch` 

https://github.com/nolanlawson/ember-pouch/blob/master/index.js#L31

Perhaps you can start there?

backspace:
@broerse, @fivetanley is the person who migrated Ember Data into an addon, allowing us all to remove it from `bower.json` :grinning:

Your ideas sound good to me. I’m particularly interested in the JSON API-like adapter idea. Making it configurable would be great so people could keep using `relational-pouch` if they want but also able to opt in to the JSON API adapter, if it has advantages for them. What are you thinking would be advantages to such an adapter?

Since you’re intimately acquainted with Ember Data, do you have any idea whether it’s possible to automate the saving of parent records in a `hasMany` relationship? So we don’t need things like [this](https://github.com/nolanlawson/ember-pouch/pull/119/files).

broerse:
Did not know :smile:. Perhaps I also misunderstand. What does the pouchdb ember-cli addon have what the pouchdb NPM version does not have?

fivetanley:
It's just the that npm version is currently more work to get in right now. :p

broerse:
@fivetanley If you help me I am willing to putt in the energy to move to the npm versions.

broerse:
@fivetanley What work needs to be done to make https://www.npmjs.com/package/pouchdb work with ember-pouch ??

nolanlawson:
[...]

PR #151, opened 2016-10-18
Switched to npm versions of pouchdb

Fed03:
This PR allow to use the npm version of pouchdb and its plugins, therefore addressing #120.
In order to achieve this goal, I had to create a tiny broccolijs plugin that creates the amd wrapper and pass it to browserify. I took this idea after long searching from the awesome [ember-browserify](https://github.com/ef4/ember-browserify), so thnx to @ef4.

I hope the code speaks for itself, but I'll be glad to explain if needed

### DIFF
--- a/addon/initializers/pouchdb-plugin.js
+++ b/addon/initializers/pouchdb-plugin.js
@@ -1,0 +1,14 @@
+import PouchDB from 'pouchdb';
+import RelationalPouch from 'relational-pouch';
+import Find from 'pouchdb-find';
+
+export function initialize() {
+  PouchDB
+    .plugin(RelationalPouch)
+    .plugin(Find);
+}
+
+export default {
+  name: 'pouchdb-plugin',
+  initialize
+};
--- a/app/initializers/pouchdb-plugin.js
+++ b/app/initializers/pouchdb-plugin.js
@@ -1,0 +1,1 @@
+export { default, initialize } from 'ember-pouch/initializers/pouchdb-plugin';
--- a/bower.json
+++ b/bower.json
@@ -3,9 +3,6 @@
   "dependencies": {
     "ember": "~2.8.0",
     "ember-cli-shims": "0.1.1",
-    "pouchdb": "^5.4.5",
-    "relational-pouch": "^1.4.4",
-    "pouchdb-find": "^0.10.3",
     "phantomjs-polyfill-object-assign": "chuckplantain/phantomjs-polyfill-object-assign"
   }
 }
--- a/index.js
+++ b/index.js
@@ -1,16 +1,23 @@
 /* jshint node: true */
 'use strict';
+const browserifyTree = require('./lib/browserify-tree');
 
 module.exports = {
   name: 'ember-pouch',
-  included(app) {
-    const bowerDir = app.bowerDirectory;
-
-    app.import(bowerDir + '/pouchdb/dist/pouchdb.js');
-    app.import(bowerDir + '/relational-pouch/dist/pouchdb.relational-pouch.js');
-    app.import(bowerDir + '/pouchdb-find/dist/pouchdb.find.js');
-    app.import('vendor/shims/pouchdb.js', {
-      exports: { 'pouchdb': [ 'default' ]}
+  treeForVendor(tree) {
+    return browserifyTree(tree, {
+      modules: [
+        {
+          module: 'pouchdb-browser',
+          resolution: 'pouchdb'
+        },
+        'relational-pouch',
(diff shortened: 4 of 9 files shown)


broerse:
@fivetanley what is needed for fastboot support after we switch to NPM with this PR?

Fed03:
@broerse I think @fivetanley is too busy XD

broerse:
@Fed03 I like the work you did here but can't oversee what the broccoli hacks to use browserify in an add-on will do for all people using ember-pouch. What will happen when ember-cli supports this by default? What will happen with fastboot? I think ember-pouch users with more knowledge should review this. See also https://github.com/ef4/ember-browserify/issues/105

Fed03:
@broerse you're right. Just two things to think about: 

1. I think the ember-cli core would move from bower to es6next packages so no CJS at all... these thing is yet to be "standardized" in the entire js community so it's my opinion it'll take long time
2. I strongly believe in the deprecation policy brought by ember-cli incremental update

As for fastboot I honestly have not played with it so I don't have any clue...
Maybe we could bother @stefanpenner and listen to what he thinks about these

broerse:
@rsutphin @aexmachina @dweremeichik @fsmanuel If you have time please take a look at this. Is it to much a hack?

broerse:
@ef4 I suggest you start using PouchDB so we can ask you questions about this PR 😄

ef4:
Maybe I don't have all the context here, but I think this can be done more simply.

If the same files are available via npm that are available via bower, you can merge the npm packages into your vendor tree and continue using `app.import` as before. Something like:

```js
  treeForVendor: function(tree){
    var pouchPath = path.dirname(require.resolve('pouchdb-browser'));
    var pouchTree = new Funnel(this.treeGenerator(pouchPath), {
      srcDir: '/',
      destDir: 'pouchdb'
    });
    return mergeTrees([tree, pouchTree]);
  },
  included() {
    app.import('vendor/pouchdb/lib/index.js');
  }
```

Alternatively, if you need to package up a dependency graph of CJS modules into coherent AMD modules, you can just invoke browserify directly:

```js
var browserify = require('broccoli-browserify');
module.exports = {
  treeForVendor: function(tree){
    var src = this.treeGenerator(path.join(require.resolve('pouchdb-browser'), '..', '..'));
    var pouchdb = browserify(src, {
      outputFile: 'browserified-pouch.js',
      require: [['./lib/index.js', {expose: 'pouchdb'}]]
    });
    return mergeTrees([highlight, tree]);
  },

  included: function(app) {
    app.import('vendor/browserified-pouch.js');
 }
}
```

dweremeichik:
@broerse @Fed03 If it can be simplified per what @ef4 said, perhaps we should give it a try? Otherwise, it does move the ball forward for the whole fastboot thing.

@broerse in regards to one of your [previous comments](https://github.com/nolanlawson/ember-pouch/pull/151#issuecomment-254699388), we would probably want to change our adapter blueprint to add a guard for the fastboot environment. What comes to mind is disabling the `localDb` when on the fastboot server, and using pouch-db as a pass-through to whatever couch instance you might be using; this way the fastboot server does not need to save state. Anything else that we would likely need to do would be [here](https://ember-fastboot.com/docs/addon-author-guide).

broerse:
@ef4 Thanks!

Fed03:
@ef4 I tried your second solution but it doesn't work, PouchDB is just a global now not an importable module:
`Could not find module 'pouchdb' imported from ...`

I'll try to find a better way

ef4:
I think if you mess with the [browserify options](https://github.com/substack/node-browserify#usage) you can make sure it generates an importable AMD module.

Fed03:
I didn't dig into it yet but I think you can generate UMD files that are not supported by ember-cli

broerse:
@Fed03 I can try the first `treeForVendor` way to skip the use for browserify. [This](https://github.com/substack/node-browserify/issues) does not comfort me ;-)

broerse:
@Fed03 @dweremeichik Something like this https://github.com/nolanlawson/ember-pouch/pull/162 seems more robust.

Fed03:
@broerse yes as long @nolanlawson keeps releasing the full browser version.
I know that every pouchdb pkg have the `jsnext:main` entry in package.json but I don't know anything about how use it....maybe we should look into [this](https://github.com/rollup/rollup/wiki/jsnext:main)

broerse:
Thanks @Fed03, @ef4 we just released the first npm only version: https://github.com/nolanlawson/ember-pouch/releases/tag/v4.2.0

Issue #148, opened 2016-10-12
queryRecord array deprication


dweremeichik:
When running the tests for ember-pouch with the `--server` flag, the following warning is issued in the chrome console:

```
DEPRECATION: The adapter returned an array for the primary data of a `queryRecord` response. This is deprecated as `queryRecord` should return a single record. [deprecation id: ds.serializer.rest.queryRecord-array-response]
```

PR #152, opened 2016-10-18
Fixes queryRecord issue #148

dweremeichik:
This fixes issue #148, and removes the following warning:

DEPRECATION: The adapter returned an array for the primary data of a `queryRecord` response. This is deprecated as `queryRecord` should return a single record. [deprecation id: ds.serializer.rest.queryRecord-array-response]

---
- The appropriate behavior is outlined in the Ember Data API, [here](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord).
- Test added to ensure that null is returned when the filter doesn't resolve a record.

Note:
- The warning appears to be in place `until: '3.0'`. 
- Since `queryRecord` was just an alias for `query` previously this would be a non-backward compatible change.


dweremeichik:
Just a note, the warnings were [added](https://github.com/emberjs/data/pull/4300) in Ember Data 2.6.0.
Build fails for 2.1 stack and below.

fsmanuel:
@dweremeichik maybe [this](https://github.com/funkensturm/ember-local-storage/blob/master/addon/adapters/base.js#L108-L113) helps to make it backwards compatible.

broerse:
@dweremeichik We started testing from Ember 2.4 and I tried in https://github.com/broerse/ember-pouch/tree/queryRecord to fix this PR but your added test still fails. Do you have time to take a look and fix your PR. I like to merge this.

dweremeichik:
@broerse haven't used ember-pouch in a while. However it used to build in 2.4 without issue. May have something to do with the tests changing when the hasmany "split tests" were added.

dweremeichik:
Also, a brief look at the logs tells me it only breaks when the tests are "sync" not sure what that feature is, added after my initial PR, must  be.

addon/adapters/pouch.js
```diff
@@ -340,7 +340,16 @@ export default DS.RESTAdapter.extend({
   },
 
   queryRecord: function(store, type, query) {
-    return this.query(store, type, query);
+    return this.query(store, type, query).then(results => {
+      let result = {};
```
jlami: You need to edit the original `results` object. Since this can have additional entries, which are sideloaded.

addon/adapters/pouch.js
```diff
@@ -340,7 +340,16 @@ export default DS.RESTAdapter.extend({
...
+      let recordType = this.getRecordTypeName(type);
+      if(results[pluralize(recordType)].length > 0){
+        result[recordType] = results[pluralize(recordType)][0];
+      } else {
+        result[recordType] = null;
+      }
```
jlami: after using the singular `recordType` to set the expected single item, you need to `delete results[pluralize(recordType)];` to remove the plural entry

broerse:
@dweremeichik this PR fixed the failing tests. https://github.com/pouchdb-community/ember-pouch/commit/86601e6a14588d14abc8538024be755eb2a5fa71

dweremeichik:
Awesome work @jlami @broerse!

PR #172, opened 2017-02-02
Create Pouch Service

hamzahio:
Good day Nolanlawson,

This pull request for creating service for the initiated db and we can access it from anywhere.

**Benefits**:
Authentication access current db and add authentication code.

stop syncing or anything else.


akatov:
This looks too restrictive. As a consumer I am managing several pouches, e.g. one local, one remote, sometimes a local memory-only pouch. I roll my own service that manages these pouches and updates the adapter's db as and when necessary.
@hamzahio I don't think something like this should be part of ember-pouch, as it's too restrictive...

broerse:
As shown by https://bloggr.exmer.com/ you need something like [this](https://github.com/broerse/ember-cli-blog/blob/master/app/adapters/application.js) and not sync to get it working. I agree with @akatov that we don't want restrictive code in ember-pouch. I need to write documentation but authentication can be done with https://github.com/martinic/ember-simple-auth-pouch 

Perhaps it is better to create ember-pouch-service but it still needs to be more flexible to support for instance the scenario above.

PR #173, opened 2017-02-03
Remove ember-getowner-polyfill

ntodd:
Ember.getOwner() has been available since Ember 2.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "ember-data": "^2.8.0",
     "ember-disable-prototype-extensions": "^1.1.0",
     "ember-export-application-global": "^1.0.5",
-    "ember-getowner-polyfill": "1.0.1",
     "ember-load-initializers": "^0.5.1",
     "ember-resolver": "^2.0.3",
     "loader.js": "^4.0.1"
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -1,6 +1,5 @@
 import Ember from 'ember';
 import DS from 'ember-data';
-import getOwner from 'ember-getowner-polyfill';
 
 import {
   extractDeleteRecord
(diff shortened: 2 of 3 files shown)


broerse:
Ember 2.0 , 2.1 and 2.2 seem to fail on this PR. i don't mind supporting Ember 2.4 and up only

ntodd:
@broerse I agree, but maybe we could make this nicer?  Could do something like: https://github.com/stefanpenner/ember-moment/blob/master/addon/utils/get-owner.js

broerse:
@ntodd No I think if we support Ember 2.4+ we will not break many projects. I was thinking to add something like [this](https://github.com/mharris717/ember-cli-pagination/blob/master/index.js) for Ember via the new `forEmber` added for `ember-source`.

If you can make a PR for this and make some new [stacks](https://github.com/nolanlawson/ember-pouch/blob/master/.travis.yml) it would be nice. I can perhaps do this on Sunday myself.

package.json
```diff
@@ -48,7 +48,6 @@
     "ember-data": "^2.8.0",
     "ember-disable-prototype-extensions": "^1.1.0",
     "ember-export-application-global": "^1.0.5",
-    "ember-getowner-polyfill": "1.0.1",
```
rwjblue: The dep here is still needed if we are intending to support Ember versions prior to 2.3, but we should update the version to `^1.1.0`.

The deprecation that is being fixed is specifically about refactoring `import getOwner from 'ember-getowner-polyfill';` to just `const { getOwner } = Ember;` (e.g. just use `Ember.getOwner`).

ntodd:
@broerse Let me know if you want to take as-is and set the minimum supported Ember version to 2.3 or if you want me to update to use the polyfill correctly.  Thanks!

broerse:
@ntodd I want to take as-is and only support 2.3+

PR #176, opened 2017-02-17
Fix sort documentation and add additional notes

ntodd:
Fixes the example code and also adds some helpful instructions that are not initially obvious.

The comment regarding index matching is pulled from https://github.com/nolanlawson/pouchdb-find/blob/master/lib/adapters/local/find/query-planner.js#L121


broerse:
Thanks!

PR #186, opened 2017-07-26
Update Ember CLI and PouchDB

ntodd:
Upgrade project to ember-cli 2.14.1 and use pouchdb 6.3.4.  Also updated tests to pass under eslint and use LTS versions of Ember CLI on travis.

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -147,7 +147,7 @@ export default DS.RESTAdapter.extend({
     if (type.documentType) {
       schemaDef['documentType'] = type.documentType;
     }
-    
+
     let config = getOwner(this).resolveRegistration('config:environment');
     let dontsavedefault = config['emberpouch'] && config['emberpouch']['dontsavehasmany'];
     // else it's new, so update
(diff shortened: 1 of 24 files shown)


backspace:
Thanks for doing this! The `ember-beta` job is failing, seemingly unable to find `components/ember-data#beta`. I tried deleting the cache on Travis to no avail. I have to leave now but I can look at it again later if no one else has ideas.

backspace:
hmm, I restarted it again and it worked this time! The canary job is still failing with a similar error to the one the beta job had, but that’s an allowed failure, so ¯\\\_(ツ)\_\/¯

broerse:
This is looking good!

broerse:
Travis is failing on a version number change in package.json and some readme change? It seems the new `trusty` is not so trusty https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming?utm_source=web&utm_medium=banner&&utm_campaign=trusty-default

Perhaps rerun travis in a few hours?

addon/adapters/pouch.js
```diff
@@ -163,40 +163,36 @@ export default DS.RESTAdapter.extend({
...
-        	if (inverse) {
-	        	if (inverse.kind === 'belongsTo') {
-	        		self.get('db').createIndex({index: { fields: ['data.' + inverse.name, '_id'] }});	
-	        		if (options.async) {
-	        			includeRel = false;
-	        		} else {
-	        			options.queryInverse = inverse.name;
-	        		}
-	        	} else {
-	        		console.warn(type.modelName + " has a relationship with name " + rel.key + " that is many to many with type " + rel.type + ". This is not supported");
```
jlami: Should these be preserved? Maybe with `Ember.warn(msg, false)`

tests/dummy/app/templates/application.hbs
```diff
@@ -0,0 +1,5 @@
+{{!-- The following component displays Ember's default welcome message. --}}
+{{welcome-page}}
```
jlami: Is this file actually needed?
Would this even work without the ember-welcome-page package?
https://github.com/ember-cli/ember-addon-output/blob/master/tests/dummy/app/templates/application.hbs doesn't have this.

Issue #188, opened 2017-07-27
Assertion error if requests are sent right after each other


tobiasWenger:
I have a problem with deleting a record just after saving it (Assertion Failed: Cannot call writableTag after the object is destroyed).

Unfortunately I could not create an Ember Twiddle because the ember-pouch addon could not be loaded. Instead I created a GitHub repo (https://github.com/tobiasWenger/pouchdb-error)

Post model:
```
comments: hasMany('comment', { async: false, dontsave: true }),
messages: hasMany('message', { async: false, dontsave: true })
```

Post controller:
```
import Ember from 'ember';

const {
  Controller,
  get,
  computed: { union },
  RSVP: { Promise }
} = Ember;

export default Controller.extend({
  commentsAndMessages: union('model.comments', 'model.messages'),

  actions: {
    addStuff() {
      this._createAndSave().then(() => {
        this._deleteAll();
      });
    },

    addStuffWithDelay() {
      this._createAndSave().then(() => {
        setTimeout(() => {
          this._deleteAll();
        }, 3000);
      });
    }
  },

  _deleteAll() {
    get(this, 'commentsAndMessages').forEach((commentOrMessage) => {
      commentOrMessage.destroyRecord();
    });
  },

  _createAndSave() {
    const store = get(this, 'store');
    const post = get(this, 'model');

    const comment = store.createRecord('comment', {
      post: post
    });

    const message = store.createRecord('message', {
      post: post
    });

    return Promise.all([
      comment.save(),
      message.save()
    ]);
  }
});
```

The two functions 'addStuff' and 'addStuffWithDelay' are doing basically the same thing. The only difference is the added timeout in 'addStuffWithDelay'.

I hope someone can point me in the right direction to solve this error.

jlami:
Have you tried ember-data ~2.13
There are some problems with 2.14

I'll see if I can repro with your repository.

tobiasWenger:
Thanks for your fast response. I just tried ember data 2.13.2 and the error is still there.

jlami:
I also seem to get the error "Attempted to handle event `pushedData` on <comment:....> while in state root.deleted.inFlight" if I mark the models as async. Maybe that one will give more results while searching, as most of Ember data is expected to work with async=true.

It looks like Ember Data is doing a findRecord or something else that pushes data to the store when it is not supposed to. Might be a race condition, but I would expect the promises to guide this in the right direction, since there is no real remote server in this example.

[...]

PR #189, opened 2017-07-27
Ignore same revision changes

jlami:
Possible solution for #188

Will look for a better solution

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -80,10 +80,11 @@ export default DS.RESTAdapter.extend({
       this.unloadedDocumentChanged(obj);
       return;
     }
-    if (!recordInStore.get('isLoaded') || recordInStore.get('hasDirtyAttributes')) {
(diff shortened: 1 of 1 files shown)


broerse:
Should I merge this?

jlami:
Not yet I think, this is a temporary fix to see if this is the way to go. I have an extra issue open on ember-data to see if this is something they can fix.

tobiasWenger:
@jlami Could you find out if it is the way to go? I would like to have this "hack" merged into master.

jlami:
@tobiasWenger Do you have experience with this hack in place. I must admit that I have not tested it as thoroughly as I wanted yet.

tobiasWenger:
@jlami My app is still in the development phase but up to now I did not encounter any problems. I added this hack locally to the newest version (4.3.0) to test it and it seams to work fine.

jlami:
Let's use this then and we will see if there are any problems or if we can even improve on this.

tobiasWenger:
@jlami Many thanks and I really appreciate all of your work. I would be lost without people like you. Keep up the good work!

PR #192, opened 2017-09-08
Eventually consistent

jlami:
See readme for further notes


Fed03:
just a quick question:
I call `this.store.findRecord('user',1)`, what happens if user with id 1 does not exist? does it hangs indefinitely?

PR #193, opened 2017-09-08
allow usage of limit parameter in pouchdb adapter queries

akatov:


broerse:
@akatov Can you add something to the readme about this?

akatov:
sure. will do this and a rebase as well

afinne:
It could be an idea to also include skip?

What we did in our project was:

```
    queryParams = {
      selector: selector
      limit: query.limit ? 100
      skip: query.skip
    }
    queryParams.sort = @_buildSort(query.sort) unless Ember.isEmpty(query.sort)

```

Note, we use coffeescript

akatov:
@afinne good idea! @broerse I'm happy to make another PR for the `skip` parameter

akatov:
see #198 for `skip` parameter

broerse:
@akatov Thanks will merge both today.

Issue #168, opened 2017-01-24
query not filtering on modelType


eneuhauser:
When I call `get` on an `async` `hasMany` relationship that isn't loaded, I do not get the related models back. To get around this, I tried calling `query` filtering on the `belongsTo` attribute, but I'm getting back additional models that also have the same `belongsTo` relationship as the model type I queried.

If I have a model structure like the following:

```javascript
const Author = Model.extend({
    name: DS.attr('string'),
    posts: DS.hasMany('post'),
    comments: DS.hasMany('comment')
});
const Post = Model.extend({
    title: DS.attr('string'),
    content: DS.attr('string'),
    author: DS.belongsTo('author')
});
const Comment = Model.extend({
    text: DS.attr('string'),
    author: DS.belongsTo('author')
});
```

With just the `author` loaded and I call `author.get('comments')`, I do not get back any results. If I call:

```javascript
store.query('comment', { filter: { author: author.get('id') }});
```

I will also get back posts for the author, but they will be loaded in the store as `comment` and the `text` attribute will be `undefined`.

Additional information, I'm using the `dontsavehasmany` flag, ember-pouch 4.2.3, and ember-data 2.10 (and 2.11.0-beta.1).

jlami:
Using query with only the 'author' as a selector is a known bug: https://github.com/nolanlawson/relational-pouch/issues/73

But the query that is used to get the entries for the hasMany should work with a better query that does restrict by type (filter on id). I'll try to see if I can reproduce this, but this should be covered by the test suite. Maybe you could build a ember twiddle?

eneuhauser:
Thanks for the quick feedback. I had searched around to see if this was an existing issue before submitting, but did not find any results. I'll try and create a Twiddle, but I don't see ember-pouch hosted on a CDN to add as a dependency.

Is it possible to build my query to filter on ID as well? I'm thinking of just adding a `type` attribute and filtering on that.

backspace:
If you use [Ember Twiddle](https://ember-twiddle.com), @eneuhauser, you can put dependencies in the `twiddle.json`, similarly to how Ember CLI dependencies work.

jlami:
You can see the code for querying by type here: https://github.com/nolanlawson/relational-pouch/blob/master/lib/index.js#L426
The _id filter on that line does what you want, though the parameters for lt and gt might be a bit confusing.

eneuhauser:
[...]

PR #194, opened 2017-09-08
filter results of pouch adapter query by correct type

akatov:
right now when using `query`, the ember-pouch adapter actually returns all document types from the pouchdb that match the selector...

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -351,7 +351,15 @@ export default DS.RESTAdapter.extend({
       queryParams.sort = this._buildSort(query.sort);
     }
 
-    return db.find(queryParams).then(pouchRes => db.rel.parseRelDocs(recordTypeName, pouchRes.docs));
+    var idType = type.documentType || recordTypeName;
+    var recordTypeNameRegex = new RegExp('^' + idType + '_[0-3]');
+
+    return db.find(queryParams)
+      .then(function({ docs }) {
+        let matchinDocs = docs.filter(({ _id }) => _id.match(recordTypeNameRegex));
+        return { docs: matchinDocs };
+      })
+      .then(pouchRes => db.rel.parseRelDocs(recordTypeName, pouchRes.docs));
   },
 
   queryRecord: function(store, type, query) {


addon/adapters/pouch.js
```diff
@@ -351,7 +351,15 @@ export default DS.RESTAdapter.extend({
       queryParams.sort = this._buildSort(query.sort);
     }
 
-    return db.find(queryParams).then(pouchRes => db.rel.parseRelDocs(recordTypeName, pouchRes.docs));
+    var idType = type.documentType || recordTypeName;
+    var recordTypeNameRegex = new RegExp('^' + idType + '_[0-3]');
```
jlami: I don't know if a regex is really the right way to go here. Feels a bit overkill. But using the normal way to parse the ID with parseDocID might also be too much. A _id.startsWith(idType + '_') might be enough here?

jlami: oh, maybe babel does not have polyfills yet for startsWith by default? So the RegExp should probably be ok. Technically the idType should be escaped, but the characters that would need escaping would probably not be correct Ember Model names

broerse:
@akatov I agree with @jlami that it is better to pass an extra parameter to the find selector. But for now this PR does the job. Thanks!

afinne:
A bit late now, but a comment nonetheless. This solution makes essentially `skip` and `limit` unreliable. You might end up with no results, even though data exists.

A solution more along what has been suggested in #168 makes more sense.

What we ended up implementing in our project is to actually have a `_typeFoo: true` property for model `Foo` and add that in as a criteria automatically when querying. 
The reason we did not go for the `_id` approach was that we got problems with index selection. I.e. the find plugin decided to use our custom index instead of the special `_all_docs` index when trying to follow an ember one-to-one ('belongsTo') relationship. This issue might have been fixed in the latest `pouchdb-find`. I haven't tried out the `_id` approach with that one.

broerse:
@afinne It is not released yet. Your point is valid. @akatov should we revert this PR?

akatov:
@broerse probably best to revert and solve this properly. Also, I hadn't seen #168...

PR #195, opened 2017-09-08
clean up db changes listener

akatov:
I had issues where in acceptance tests an adapter from the previous run would fire on changes to the database of an adapter from the current run...


jlami:
That sounds like an excellent catch!

broerse:
@jlami Do you want to merge this in beta.3 ?

jlami:
This can go in beta, or in master, either way is fine.

broerse:
@akatov Thanks! We will release a breaking 5.0.0 but improved version soon.

PR #200, opened 2017-09-25
Added test-waiter to fix sporadic error during testing

Fed03:
During my acceptance tests I had sporadic errors like the followings:

- `Error: Assertion Failed: You can only unload a record which is not inFlight`
- `Error: Called stop() outside of a test context at Object.extend.stop`

I've discovered these errors can happen when you try to create, modify and then delete a record. In detail this happens to me caouse in my acceptance tests I destroy the db in `beforeEach`.

Some attempts have been made to resolve this odd situation, like [ember-cli-test-model-waiter](https://github.com/gabrielgrant/ember-cli-test-model-waiter) but they're buggy and they often don't work.
Trying to fix this issue I've found a similar issue on [emberfire#312](https://github.com/firebase/emberfire/issues/312) then fixed in [#344](https://github.com/firebase/emberfire/pull/344).

This PR includes the emberfire fixes, resolving the above mentioned issues.

P.S. sadly I really don't know how to test this fix, cause the issues are sporadic and unpredictable

### DIFF
--- a/addon/adapters/pouch.js
+++ b/addon/adapters/pouch.js
@@ -327,40 +327,25 @@ export default DS.RESTAdapter.extend({
   },
 
   findAll: function(store, type /*, sinceToken */) {
-    this.operationsWaiter.incrementPendingOps();
     // TODO: use sinceToken
     this._init(store, type);
-    return this.get('db').rel.find(this.getRecordTypeName(type)).then(results => {
-      this.operationsWaiter.decrementPendingOps();
-      return results;
-    }).catch(() => {
-      this.operationsWaiter.decrementPendingOps();
-      return RSVP.reject(...arguments);
+    return this._dbOperation(() => {
+      return this.get('db').rel.find(this.getRecordTypeName(type));
     });
   },
 
   findMany: function(store, type, ids) {
-    this.operationsWaiter.incrementPendingOps();
     this._init(store, type);
-    return this.get('db').rel.find(this.getRecordTypeName(type), ids).then(results => {
-      this.operationsWaiter.decrementPendingOps();
-      return results;
-    }).catch(() => {
-      this.operationsWaiter.decrementPendingOps();
-      return RSVP.reject(...arguments);
+    return this._dbOperation(() => {
+      return this.get('db').rel.find(this.getRecordTypeName(type), ids);
     });
   },
 
   findHasMany: function(store, record, link, rel) {
     let inverse = record.type.inverseFor(rel.key, store);
     if (inverse && inverse.kind === 'belongsTo') {
-      this.operationsWaiter.incrementPendingOps();
-      return this.get('db').rel.findHasMany(camelize(rel.type), inverse.name, record.id).then(results => {
-        this.operationsWaiter.decrementPendingOps();
-        return results;
-      }).catch(() => {
-        this.operationsWaiter.decrementPendingOps();
-        return RSVP.reject(...arguments);
+      return this._dbOperation(() => {
+        return this.get('db').rel.findHasMany(camelize(rel.type), inverse.name, record.id);
       });
     } else {
       let result = {};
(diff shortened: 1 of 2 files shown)


jlami:
I personally don't really like the idea of fixing problems in other packages here. But I do feel your pain in the inconsistencies in testing. I will think about it a little bit and we'll talk tomorrow @broerse...

Fed03:
@jlami what do you mean? what pkgs?

broerse:
@Fed03 doesn't the Eventually consistency in v5.0.0-beta.2 solve this somehow automatically? Did you see the tests fail with v5.0.0?

jlami:
Maybe #195 actually helps here?

@Fed03 If I'm playing devils advocate, I think that the problems arise from how the test environment works. Jumping through hoops in ember-pouch might solve it, but it does not feel like the correct way to fix things. I'll have to look further into the problems to know whether this is correct, or if it is really something we should fix.

But the proposed fix feels like overkill. Maybe I'm missing something, but it looks like you only increment and decrement a counter for every promise we have. Shouldn't the test environment or ember-data be able to handle these hanging promises themselves? Something like ember-cli-test-model-waiter seems like the way to go. But this does not fix everything you say? Then maybe we (or the test environment) should look into that?

Maybe it would be nice to know if something like this repro would trigger the error here too: https://github.com/firebase/emberfire/issues/312#issuecomment-143109856

Fed03:
I think that ember tries to handle the promises but seems that not all promises or run loops are included by default when determining what makes the test helpers wait, so we have to make them explicit using `registerTestWaiter()`

However I'm going to try to reproduce the error and what happens with the unmodified 5.0.0

Fed03:
I've created a repo that mirrors [https://github.com/firebase/emberfire/issues/312#issuecomment-143109856](url).
You can find it [here](https://github.com/Fed03/sporadic-errors). It provides 3 tags:

1. v4.3.0 - It uses the last stable version of ember pouch - **FAILING**
2. v5.0.0 - It uses the last beta version of ember pouch - **FAILING**
2. v5.0.0 - It uses my waiter version of ember pouch - **PASS**

jlami:
Thank you, I will look into it later today.

Fed03:
I can make the solution a little bit more elegant if gets approved

Il mar 26 set 2017, 12:37 jlami <notifications@github.com> ha scritto:

broerse:
@Fed03 Looking closer at this I don't think we should change the adapter to help tests like this. That said I still think we should fix this. In the current form I don't think we should approve this.

jlami:
I believe the test `afterEach` does a application teardown, that should stop anything from going wrong. But I want to investigate what kind of change or promise triggers these errors.

Fed03:
@jlami @broerse let me know if I can help in any ways.

jlami:
I think your problem is due to an uncaptured promise. If I change the way the test navigates I can get it to work as expected. I removed the transitionToRoute in the action and added an explicit 
```
  andThen(function() {
  	visit("/view-user");
  });
```
before the original andThen in the test.

I believe this way the explicit `visit` registers the Promise in the `route.model`. But it is unfortunate that the `transitionToRoute` in the controller does not work this way.

I haven't completely ruled out that my change is actually capturing the model promise, so I will continue debugging. But my first impression is that embers test helpers for click are not not completely working as we would like.

Fed03:
I think it's worth to note that this problem arises also when using `async/await` instead of promises and with `ember-native-dom-helpers` that replaces `click, visit, fillIn, find...`

jlami:
I'm having some trouble looking into the click helper. Chrome does not seem to want to step into it. And Firefox does not seem to show the same out of order execution.
I will see if I can find out what the problem is, but at this point I blame the `click` handler not capturing the promise that is started in the action handler.

Fed03:
Do you think the problem is within the [`wait()`](https://github.com/emberjs/ember.js/blob/v2.15.0/packages/ember-testing/lib/helpers/wait.js#L38) function?

jlami:
Maybe. I'm now debugging again in Chrome, with source maps disabled :'( And it might be that the wait function is not aware of the transition that the action started. But I don't quite see how the visit does make this work though. I'll keep debugging.

Fed03:
Revolutional idea: wht if we wrap the promises inside run loop?

jlami:
I would have expected ember-testing to actually do exactly that. But as far as I can tell, they only make a TestingPromise for use inside the helpers. But really wrapping all the promises might be too much. I think ember should be aware of promises returned by action handlers

I think `visit` sets https://github.com/emberjs/ember.js/blob/v2.15.0/packages/ember-testing/lib/helpers/wait.js#L45 `activeTransition`
While actions that are triggered with `visit` should do the same, but `wait` is called before that can happen. And I don't see any way the testing helpers really wrap the promises an action returns.

broerse:
@Fed03 I know ember-pouch is the only place where you can fix this until Ember is fixed but is needs to fixed there. @jlami will create a bug-report/issue on Ember itself. It almost needs a RFC. I will close this PR.

jlami:
I've been talking to alexspeller and some others on Slack and I don't know if there is any traction on getting this into ember core. But I did get some pointers to get this to work with addons.

As a starter https://github.com/bendemboski/ember-concurrency-test-waiter might be good way to say that an action is actually testable. While I don't really like the hard dependency on ember-concurrency, I could overlook that. But for now the explicit way you have to call `withTestWaiter` also signals to me that this is not really the way to go.

I ideally would like some way to make actions waitable without having to change application code. This way even addons that know nothing about concurrency could be used.

They also pointed me towards https://github.com/kellyselden/ember-wait-for-test-helper/ which makes it easier to make custom waiters. But this still does not allow to wait for actions, as ember does not have a way to track the results of action handlers.

So after that, I started fiddling with a twiddle that I used on Slack to explain the issue and got it to work somewhat: https://ember-twiddle.com/6714d1ff03d0d1893d0bea6e571ae0d3?openFiles=tests.acceptance.wait-for-action-test.js%2C&route=%2Fshould-endup-here
There is a override of the normal way ember does `ActionHandler.send` on initializers/action-waiter.js. There I also create an extra `waitForActions()` test helper that can be used to explicitly wait for the actions to settle.

I still have some work to do, because it does not support bubbling right now and requires some more testing. And I don't know if this will work with component actions yet. So still more work to do, but it does look promising :)
The amount of hacking required to make this work does suggest to me that this might actually be something that ember-core would be better suited for. But for now such a hack could be enough.

Fed03:
I don't know how to thank you for such researches! I didn't think this issue could be so complicated to fix...
As always let me know if I can be of any help.

Ps. could be useful to pull some ember core members in the discussion?

jlami:
Don't really want to bother them right away, since alexspeller did mention he didn't really see this changing, but eventually it might be a good idea.
And it is actually really fun to dig around in core, so don't worry about it. I might call upon you if I do make an addon for this, since you are actually the one using it ;)

PR #187, opened 2017-07-27
Fix some jobs on Travis Trusty

backspace:
This is a pretty hackish solution to [this problem](https://github.com/ember-cli/ember-try/issues/125), also discussed [here](https://github.com/ember-cli/ember-try/issues/133): `ember-try` uses the presence of a `yarn` command to determine whether to `yarn install`. Something was [going wrong with the `yarn install`](https://travis-ci.org/pouchdb-community/ember-pouch/jobs/257985277#L3444-L3454) on the `ember-release` and `ember-beta` scenarios.

This fixes those scenarios but then breaks the ones before them, the inverse of what was happening!

I’m going to ask over on the `ember-try` thread about whether it’s possible to bypass Yarn more elegantly. I’ll also see if I can override Ember CLI’s built-in `ember-try` version with an older one that doesn’t use Yarn.

PR #201, opened 2017-11-10
Restore ember-try configuration for Ember Data

backspace:
Thanks to @jlami for [pointing out](https://github.com/pouchdb-community/ember-pouch/pull/187#issuecomment-343308963) that I mistakenly deleted this.

I haven’t run the tests locally, I wouldn’t be surprised if this still needs some tweaking. We’ll see!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
   matrix:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
+    - EMBER_TRY_SCENARIO=ember-2.4-stack
     - EMBER_TRY_SCENARIO=ember-lts-2.8
     - EMBER_TRY_SCENARIO=ember-lts-2.12
     - EMBER_TRY_SCENARIO=ember-release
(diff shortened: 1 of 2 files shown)


backspace:
The blueprint has both `ember-release` and `ember-default`, maybe this is overkill for us 🤔

broerse:
We will go with it for now. Thanks!

jlami:
This looks much better! Thanks for your work @backspace 

The only thing I can see is that the 'canary' scenario is failing in travis. Npm does not have a 'canary' tag for ember-data, so travis is failing on the `npm install`. Maybe we could use the bower way there?  But https://github.com/components/ember-data is also in a confusing state and does not have a canary tag/branch as far as I can see. So maybe just use the beta there?

https://github.com/components/ember-data/pull/38#issuecomment-269775605 seems to indicate that we could use the github master branch? But I always thought that github could not be used directly, since the publish step would generate the distribution packages. But maybe this is only for ember-source as ember-data is a 'normal' addon?

ember-default should be the one we have setup while developing, while the ember-release could easily change to 2.17 in the coming days. So I think it is good to have them both.

PR #202, opened 2017-11-11
Fix Ember Data canary ember-try scenario

backspace:


backspace:
I don’t mind that much because we’re only needing to use it for tests, with no actual `bower.json` in the project 😆

jlami:
Yeah, that is true. But I now @broerse would like to have no mention of bower at all :wink:
But this will have to do for now :smile:

PR #209, opened 2018-01-14
#207 Update documentation about Offline First

### LINKED ISSUE #207: From AppCache to Service Workers
At the moment [we're suggesting](https://github.com/pouchdb-community/ember-pouch#offline-first) users to use [broccoli-manifest](https://github.com/racido/broccoli-manifest) to setup an Offline First app using Ember Pouch.

But the application cache has been [removed from the Web standards](https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache). I have been using [Ember Service Worker](http://ember-service-worker.com) instead and it seems to be the way to go for future-proof-offline-first-web-apps.

Would the team be interested if I'd update the Offline-First section using Ember Service Worker?

Btw, thanks for everyone's work on this extremely useful plugin 👏 🌻

Voles:
Fixes #207.


broerse:
Thanks!

Issue #229, opened 2018-09-03
Cannot get attachments working (error in serializer)


bastientanesie:
Hi there!

I'm having a hard time working with `ember-pouch`  (v6.0.0) and attachments on a `ember-cli` v3.3 app.

According to the [README](https://github.com/pouchdb-community/ember-pouch/blob/v6.0.0/README.md#attachments), I've setup my model like [this (stripped down) gist](https://gist.github.com/bastientanesie/e4ccdeab19f9cf57967d6377b1dffe65).

The user selects a file with the `<input type="file">`, I get the `File` object in the controller and then add it to the `User` model as the README says (here as a Blob, but I also tried the base-64 way).

That throws me an error in the [Serializer](https://github.com/pouchdb-community/ember-pouch/blob/v6.0.0/addon/serializers/pouch.js#L71): *Uncaught TypeError: Cannot convert undefined or null to object*.  

```javascript
extractAttributes(modelClass, resourceHash) {
  let attributes = this._super(modelClass, resourceHash);
  let modelAttrs = get(modelClass, 'attributes');
  modelClass.eachTransformedAttribute(key => {
    let attribute = modelAttrs.get(key);
    if (this._isAttachment(attribute)) {
      let fileNames = keys(attributes[key]); // <----------- it breaks here
      fileNames.forEach(fileName => {
        attributes[key][fileName] = resourceHash.attachments[fileName];
      });
    }
  });
  return attributes;
},
```

When the error throws, `fileNames` is `undefined`, `attributes` looks like this:
```javascript
{
  email: "bastien@wixiweb.fr",
  username:  "bastien2",
  avatar: []
}
```

In CouchDB, the user's data look like that:
```json
{
  "_id": "user_2_XXXXXXXX",
  "_rev": "X-XXXXX",
  "data": {
    "email": "bastien@wixiweb.fr",
    "username": "bastien2",
    "avatar": []
  }
}
```
I've added the `avatar` field mysel, but it don't work without it anyway.

Everything looks ok, so I don't get why it breaks there. And I was unable to find any kind of help about that so here I am...

Let me know if you need more info!

jlami:
Could we see your model file? At first glance it looks like you have `attachment` there (as stated in the gist), where ember-pouch expects `attachments` (found in the readme). But I'm not sure, have not worked with them myself yet.

bastientanesie:
[...]

PR #230, opened 2018-09-04
attachment transform did not work with null

jlami:
Can help with #229


broerse:
@bastientanesie @jlami Does this need a new version now?

jlami:
Don't think this is a breaking change, but let's wait until @bastientanesie can confirm this actually works.

PR #242, opened 2019-05-18
Add testing for LTS versions up to 3.8

backspace:
I’m looking into what version of Ember and/or Ember Data is when [this test started failing](https://travis-ci.org/pouchdb-community/ember-pouch/jobs/522967859#L4453-L4458). I figured we might as well be testing the more recent LTSs.

Based on this, it looks like the problem is somewhere after 3.4, up to and including 3.8. I’ll be looking into it more deeply and maybe I can come up with a PR, but it’s daunting.


backspace:
It appears to be Ember Data 3.5.0 where the `pineapple,pineapple` thing started happening 🧐

broerse:
@backspace @jlami Updated to PouchDB 7.1.1 so now we need to fix the ember-data 3.5+ errors to release a new version.

PR #247, opened 2019-11-11
Update test matrix to include 2.18, 3.4, 3.8, and 3.12

jacobq:
Cherry-picked from #246 

Broadens test support matrix (Closes #242) and fixes test suite race condition with `store.unloadAll` that starts appearing in these newer versions (Closes #243).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "6"
+  - "8"
 
 sudo: required
 dist: trusty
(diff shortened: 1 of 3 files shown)


backspace:
hmm these failures on Travis seem like they could be due to the old Node version…?

jacobq:
> hmm these failures on Travis seem like they could be due to the old Node version…?

Oops, yes, that's due to the previous upgrade to PouchDB 7.1.1 (not me). That was a breaking change. I forced pushed and noted this.

jacobq:
I'm guessing that the [subsequent failure](https://travis-ci.org/pouchdb-community/ember-pouch/jobs/610568518#L780) has to do with the outdated `.travis.yml` inadvertently downgrading `npm` to v4.x. -- just force pushed again to see.

backspace:
Excellent, thanks for the followthrough!

PR #257, opened 2020-04-19
Fix Adapter Blueprint for Ember Octane

adezegher:
Seems when ember-pouch was updated for Ember Octane the old blue print was not correctly removed. This PR should fix this.


broerse:
Thanks!

PR #259, opened 2020-05-04
Add support for fetching non-saved hasMany relationships

adezegher:
When creating a ``hasMany`` - ``hasMany`` relationship it is possible to only save one side of the relation, to reduce the risk of merge conflict. 

However, the current implementation, only loads the `unsaved` side for a ``hasMany`` - ``belongsTo`` relationship, but not for a ``hasMany`` - ``hasMany`` relationship.

This change, allows to load the unsaved side in a ``hasMany`` - ``hasMany`` relationship, in a similar way ``hasMany`` - ``belongsTo`` does


broerse:
Not sure this is the way to go. It will not work without setting `save: true` on one side in your Ember model. If you later want to do this the other way around you are in for a challenge. We should solve this in a more constant way. What if App1 from team 1 has the 'save: true' defined on the other side then App2 from team 2?

Issue #258, opened 2020-04-30
query returns different types of models, instead of single one


adezegher:
When executing a query for a specific model, the query can return a result with other models than specified.

**Example:**

Following is expected to only return models of type ``house`` **but will actually return data of both models** ``house`` and ``person``:
```javascript
this.store.query('house',  { filter: { title: {$regex: ".*"} } });
```

**Models:**

```javascript

// app/models/house.js
import { attr } from '@ember-data/model';
import { Model } from 'ember-pouch';

export default class HouseModel extends Model {
  @attr('string') title;

}

// app/models/person.js
import { attr } from '@ember-data/model';
import { Model } from 'ember-pouch';

export default class PersonModel extends Model {
  @attr('string') title;
}
```

**Environment**
```
DEBUG: Ember           : 3.17.3
DEBUG: Ember Data      : 3.17.0
DEBUG: Ember Pouch     : 7.0.0
```

backspace:
Thanks for pointing this out. Did you have any success with addressing this in these linked commits, @adezegher? It would be preferable to include a test for this any fix, too.

adezegher:
@backspace the fix in PR #260 worked fine for me. I got pulled into something else so never got to fully debug why the tests were failing.

PR #260, opened 2020-05-05
#258: Fixes Query should only return models of given type

adezegher:


adezegher:
Not really sure why test ``can query multi-field queries`` is failing. @broerse any idea?

PR #262, opened 2020-06-29
Update README application adapter example to use JS classes

phantomphildius:
I recently added this package to a new project and had some trouble getting the app to load and found the following error in the console: `Error: Please set the `db` property on the adapter.`. After changing the setup to use class properties instead it worked!


backspace:
Nice fix, thank you!

PR #267, opened 2025-06-27
Ember 4.0

andsmedeiros:
This commit provides support for Ember and EmberData 4.x.

# What was done?

Besides resolving most deprecations and updating dependencies, this is a basis for the modernization of the codebase. 
I attempted to migrate most old JS syntax to modern alternatives while maintaining functionality. Also, old Ember and Data paradigms were mostly shifted out, like EmberObject being ditched where possible. Some quirks and hacks for obsolete Ember/Data versions were removed.

# What was not done?

I left out updating the README or the package version as that should be better handled by the package manager probably.
Also, this does *NOT* attempt to phase out EmberData's adapter paradigm yet.

# Testing

I updated all the testing code also.  However, I cannot get tests to run with scenarios `ember-lts-4.8` and `ember-lts-4.12` due to an error:

```
Global error: Uncaught TypeError: Class extends value undefined is not a constructor
```

This comes off some EmberData's private imports (IIRC `RecordArray` being undefined), but I couldn't spare the time to properly debug it.

Other scenarios run fine.

# Breaking changes

##  TL;DR

Before:
```js
import { Adapter } from 'ember-pouch';
import { PouchDB } from 'ember-pouch/pouchdb';

const db = new PouchDB('my-database');

export default class ApplicationAdapter extends Adapter {
    db = db;
}
```

After:

```js
import { Adapter } from 'ember-pouch';
import { PouchDB } from 'ember-pouch/pouchdb';

const db = new PouchDB('my-database');

export default class ApplicationAdapter extends Adapter {
    constructor(owner) {
        super(owner, db);
    }
}
```

Because `PouchAdapter` relied on `init()` to initialize its change listener, `db` could be set as a class property, as that would be set before `init()` runs. Now, the change listener must be set on the adapter constructor and keeeping `db ` as a class property would mean setting it only after the constructor runs.

The adapter constructor have been now directly defined and takes the database object as parameter. This way, it can set the property itself and correctly set the change listener during construction.

# Afterword

I will also open another PR with a branch aiming Ember and Data >=5 (including 6.0). I am available if anything is needed to getting this working well. 

🖖🏽


backspace:
Thanks for all your work on this! I looked over the PR and all seems well, I’m not sure how we should proceed, this library hasn’t seen much activity in a while and I don’t think we have conventions for maintainers to follow.

The lint job failed but based on the error it’s probably a Babel/transpiling thing 🤔

andsmedeiros:
Thanks for the quick response!
I forgot to update the GH actions, so it was installing Node 14 and that doesn't support `??=' and some other syntax.
I also fixed some more blueprint stuff that went by unnoticed.

About the release plans, I think probably we should release two majors, one aiming v4 and other v5+.
Providing a working version to v4 is meaningful for ensuring there is an upgrade path for Ember 3.28 apps (that's actually my case).

broerse:
As numerous projects are running Ember 3 and 4 with Ember-Pouch 8.0.0-beta.2, and it is evident that Ember Data 3 is still used in Ember 4 projects, like for instance, LinkedIn uses Ember 4.8.6 and Ember Data 3.28.3 at the moment, indicating that Ember-Pouch should be compatible with both Ember 3 and 4.

I am all for innovation like this, but I am not sure about the breaking changes. Since Ember-Pouch 8.0.0-beta.2 is currently used in Ember 4 projects, would it not be better to skip this innovation for Ember 4 and maintain compatibility with Ember 3? Maybe we can release the current Beta as 8.0.0 and make a breaking 9.0.0-beta.1 Ember 5+ with the Ember 5.0 #268 PR?

andsmedeiros:
@broerse TBH I was under the impression that 8.0.0-beta.2 was not working as intended. Anyway, if it works, I don't see a problem in releasing that as major and skipping this PR, going straight to https://github.com/pouchdb-community/ember-pouch/pull/268 as the next major.

broerse:
@andsmedeiros See: https://github.com/broerse/ember-cli-blog . If we release 9.0.0-beta.1 for Ember 5+ It would be super if you have time to do a PR on ember-cli-blog to bring it to the newest Ember just to demo it.

PR #268, opened 2025-06-28
Ember 5.0

andsmedeiros:
This commit provides support for Ember and EmberData >=5.x (including 6.x).

# What was done?

This PR takes the [Ember v4.x branch](https://github.com/pouchdb-community/ember-pouch/pull/267) code and adapts it to Ember/Data >=5.x.
Also, converts this into a V2 addon.

# What was not done?

This does *NOT* attempt to phase out EmberData's adapter paradigm yet.

# Testing

Tests run fine via `ember test`, but [`ember-try` fails](https://github.com/ember-cli/ember-try/issues/1108).

# Breaking changes

## IMPORTANT! Underlying model naming

Due to compatibility reasons, `ember-pouch` used camelCase model names everywhere. This was deprecated and is planned for removal at Data 6.x.
This also removes all compatibility behaviour and passes model names as-is. This means that old data may be inaccessible.
The old behaviour can be restored by overriding  the derived adapter's `getRecordTypeName()` method.

## Transform exports

If `transform:attachment` or `transform:attachments` are needed, they now need to be explicitly exported by the consuming app:

E.g., for `transform:attachment`:

```js
// my-app/app/transforms/attachment.js
export { default } from 'ember-pouch/transforms/attachment';
```

# Afterword

As said before, I am available if anything is needed to getting this working well. 

🖖🏽


backspace:
Thanks for your ongoing work on this. I don’t think I have release permissions which does seem like something to figure out before merging your other branch and then this. But I appreciate your dedication and I’ll review this in the next few days.

andsmedeiros:
Thanks @backspace! I can work on my app with this as it is currently by adding a git dependency directly. Also, the codebase is already quite useable as I am finding out, so I'm not in a hurry. 
When you can advance with this I can help further. Until then, if I catch any errors, I'll go on updating both branches.

andsmedeiros:
Note on the failing `Floating Dependencies` scenario:
I believe this is failing because, since Data 5.5, the configuration is expected to be under WarpDrive's namespace ([related to this deprecation](https://deprecations.emberjs.com/id/warp-drive-deprecate-tracking-package) and [this package](https://api.emberjs.com/ember-data/release/modules/@warp-drive%2Fbuild-config)).
I'm not sure how to work around this, maybe `@embroider/macros` can be used in `ember-cli-build.js` to conditionally define configuration.

broerse:
@jlami can you take a look at the failing checks?

